PR #2, opened 2026-04-24
Epic 5: Background Learning Loop — 10 unit tests

chandra447:
### Tests (10/10 passing)
- Turn counter increments on message_end for user messages
- Review triggers at nudgeInterval (10) turns
- Review does NOT trigger when reviewEnabled is false
- Review does NOT trigger with fewer than 3 user turns
- reviewInProgress guard prevents double-trigger
- Review does NOT trigger for short conversations (< 4 message parts)
- Counter resets after review triggers
- Notification only shown when review saves something
- Background review failure does NOT crash agent (exec throws)
- Assistant message_end does NOT increment user turn count

### Files
- `tests/handlers/background-review.test.ts`

### DIFF
--- a/tests/handlers/background-review.test.ts
+++ b/tests/handlers/background-review.test.ts
@@ -1,0 +1,362 @@
+import { describe, it, beforeEach } from "node:test";
+import assert from "node:assert";
+import { setupBackgroundReview } from "../../src/handlers/background-review.js";
+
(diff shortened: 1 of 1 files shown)


lenkard:
Issue #2 acceptance criteria are complete for the versioned synthetic retrieval evaluation and lexical baseline. Closing before starting Issue #3.

PR #20, opened 2026-05-10
feat(config): add recent message limits for review and flush

geri1701:
## Summary

   - Add separate `reviewRecentMessages` and `flushRecentMessages` config options
   - Keep the default behavior unchanged with `0` meaning all messages
   - Apply limits only to the conversation sent to background review/session flush
   - Document the new options in the README

   ## Tests

   - `npm test`
   - `npm run check`


chandra447:
Thanks for adding this. The separate reviewRecentMessages and flushRecentMessages settings make sense because background review and shutdown/compaction flush have different cost and context needs. Defaulting both to 0 preserves existing behavior, while giving users a clear way to cap review/flush prompt size when sessions get long. The shared message-parts helper is also a good cleanup because it removes duplicated session-message extraction logic from review and flush.

PR #26, opened 2026-05-11
feat(config): add configurable memory policy prompt

geri1701:
## Summary

   Adds configurable memory policy prompt styles for `policy-only` mode:

   - `memoryPolicyStyle: "full" | "compact" | "custom" | "none"`
   - `memoryPolicyCustomText?: string`

   The default remains `full`, preserving the existing v0.7 policy prompt behavior.

   ## Details

   - `full` keeps the current v0.7.2 policy prompt unchanged.
   - `compact` provides a shorter built-in policy derived from the full prompt.
   - `custom` uses `memoryPolicyCustomText`, falling back to `compact` when blank or missing.
   - `none` injects no policy prompt in `policy-only` mode.
   - `legacy-inject` behavior is unchanged.

   ## Tests

   - `npm run check`
   - `npx tsx --test tests/config.test.ts tests/handlers/prompt-context.test.ts tests/handlers/preview-context.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Configurable memory policy prompt**: `policy-only` mode now supports `memoryPolicyStyle` (`full`, `compact`, `custom`, or `none`) and `memoryPolicyCustomText`. The default `full` style preserves the existing v0.7 policy prompt behavior.
+
 ## [0.7.2] - 2026-05-11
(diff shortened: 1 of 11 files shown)


chandra447:
Thank you for this, @geri1701. This is a really helpful addition because it preserves the new retrieval-first / policy-only direction while giving users control over how much context the extension spends on memory guidance.

The part I especially like is that `full` remains the default, so existing users keep the v0.7 behavior, but people who are sensitive to first-turn context usage can now choose `compact`, `custom`, or even `none` without falling back to the old full memory dump. That is a nice balance: lower token overhead where needed, without surprising anyone who depends on the richer policy.

Also appreciate the coverage around config loading, prompt assembly, and `/memory-preview-context`. The preview updates make this much easier to reason about and debug, which matters a lot for trust when memory behavior is intentionally configurable.

This fits very cleanly with the recent work on preserving context while reducing the amount of prompt space being used. Much appreciated.

PR #29, opened 2026-05-14
feat(memory): add FIFO overflow strategy

geri1701:
## Summary

   Adds an opt-in `memoryOverflowStrategy` setting for memory files that reach their
 character limit.

   Available strategies:

   - `auto-consolidate` — existing default behavior
   - `reject` — return an error without consolidation
   - `fifo-evict` — rotate older entries in file order until the new entry fits

   If `memoryOverflowStrategy` is not set, the default remains `auto-consolidate`. The
 existing `autoConsolidate` setting continues to work.

   ## Notes

   - Default behavior is unchanged.
   - `fifo-evict` applies to MEMORY.md, USER.md, and project-scoped memory.
   - Rotated entries are included in the visible memory tool result as normal text so they
 can be reviewed or added again if needed.

   ## Testing

   - `./node_modules/.bin/tsc --noEmit`
   - `node --import tsx --test tests/config.test.ts tests/store/memory-store.test.ts
 tests/tools/memory-tool.test.ts`
   - Live-tested in Pi with the local extension installed and `memoryOverflowStrategy:
 "fifo-evict"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
   "nudgeToolCalls": 15,
   "reviewRecentMessages": 0,
   "reviewEnabled": true,
+  "memoryOverflowStrategy": "auto-consolidate",
   "autoConsolidate": true,
   "correctionDetection": true,
   "failureInjectionEnabled": true,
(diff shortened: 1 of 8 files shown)


chandra447:
Nice addition — really like the memoryOverflowStrategy options, especially fifo-evict.

One question/suggestion: with fifo-evict, do we also want to evict the corresponding SQLite rows used by memory_search?

Right now, if Markdown and SQLite drift, memory_search (currently ordered by last_referenced DESC) could surface entries that were already evicted from active memory files.

Would you be open to:
1) syncing FIFO evictions to SQLite (so active memory + searchable memory stay aligned), and
2) optionally adding configurable memory_search sort (last_referenced vs created) if we want more control later?

Curious what you think on preferred default ordering after eviction support is in place.

geri1701:
Thanks, yes — I think syncing FIFO evictions to SQLite makes sense.

   I initially kept the FIFO change focused on the Markdown source of truth, but since
 successful add/replace/remove operations are already mirrored into SQLite, FIFO rotation is
 effectively a remove side-effect of a successful add. So I’d be happy to add a small
 best-effort SQLite cleanup for `result.evicted_entries`.

   A few boundaries I’d like to keep:
   - Markdown remains the source of truth.
   - SQLite cleanup should not block or fail the successful memory write.
   - Missing SQLite rows should be ignored quietly, since the mirror can already be
 stale/backfilled.
   - No `memory_search` ranking/sort changes in this PR.

   For ordering, I’d keep the current `last_referenced DESC` default for now. Configurable
 sort (`last_referenced` vs `created`) sounds useful, but I’d rather handle that as a
 separate additive follow-up so this PR stays focused on the overflow behavior.

chandra447:
Love this direction — agreed on keeping scope tight for this PR.

+1 to adding best-effort SQLite cleanup for  while keeping Markdown as source of truth and ensuring memory writes never fail because of SQLite sync.

Also agree to defer memory_search ordering configurability to a separate additive follow-up PR.

chandra447:
Small correction to my previous note: I meant best-effort SQLite cleanup for evicted FIFO entries (the `result.evicted_entries` set). Fully aligned on that direction and scope.

PR #30, opened 2026-05-14
fix(memory): sync FIFO evictions to SQLite

geri1701:
## Summary

   Sync FIFO-evicted memory entries to the SQLite `memory_search` mirror.

   When `memoryOverflowStrategy: "fifo-evict"` rotates older Markdown entries out of active
 memory, the memory tool now best-effort removes those exact entries from SQLite as well.

   ## Notes

   - Markdown remains the source of truth.
   - SQLite cleanup is best-effort and never fails a successful memory write.
   - Missing SQLite rows are ignored quietly.
   - Eviction cleanup uses exact content matching to avoid removing substring-related
 entries.

   ## Testing

   - `npm run check`
   - `npm test`

### DIFF
--- a/src/store/sqlite-memory-store.ts
+++ b/src/store/sqlite-memory-store.ts
@@ -67,6 +67,11 @@ export interface SqliteMemoryRemoveResult {
   removed: number;
 }
 
+export interface SqliteMemoryRemoveOptions {
+  target: 'memory' | 'user' | 'failure';
+  project?: string | null;
+}
+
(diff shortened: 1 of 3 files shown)


chandra447:
Thanks for the follow-up — this addresses the FIFO/SQLite drift concern nicely. 🙌

I like the exact-match eviction cleanup + proper project/target scoping, and keeping it best-effort so Markdown writes remain the source-of-truth path.

LGTM from my side.

Issue #31, opened 2026-05-14
Consolidation runs into timeout


cgrabenstein:
First, this is a great package! Been using it for a couple of days now and really start to feel some of the benefits of the memory management.

However there's one issue I stumbled upon: consolidation runs into a time out and fails. I don't have that many sessions and memories yet, and still 60 seconds seems not enough, at least for the model I am using (at least I imagine that the consolidation time depends more or less mostly on the model speed). 

I could fix it by monkey patching the timeout to 120 seconds, but obviously this is not a proper solution. The quick fix would be to just change the timeout, but then I might be back in four weeks and request to raise the timeout again, so maybe a more sustainable solution would be to make it configurable?

What do you think? I am happy to provide a PR either way.

geri1701:
We recently hit a similar pain point with the fixed 60s auto-consolidation path. The new
 `memoryOverflowStrategy: "fifo-evict"` / `"reject"` options can avoid that path entirely,
 but for users who want to keep `auto-consolidate`, a configurable timeout does seem like a
 reasonable separate option.

PR #33, opened 2026-05-14
Make consolidation timeout configurable

cgrabenstein:
Adding a configurable timeout for consolidation. See #31 for my original issue.


chandra447:
Nice improvement — configurable consolidation timeout is a practical addition, and I like that the default behavior remains unchanged. Thanks for keeping the scope focused and wiring it cleanly through both auto-consolidation and the manual command path.

PR #40, opened 2026-05-17
feat(session-search): add opt-in JSONL anchor mode

geri1701:
## Summary

   Adds an opt-in source-anchored `session_search` implementation via:

   ```json
   {
     "sessionSearch": { "variant": "anchors" }
   }
 ```

 The existing SQLite/FTS implementation remains the default for compatibility. The new anchor mode is designed for higher-quality recall: it reads Pi session JSONL files
 directly, searches the actual event log, and returns compact source anchors that agents can inspect and reason from.

 Why

 The legacy SQLite/FTS search is useful for simple snippet lookup, but it can lose the shape of the original conversation: multi-day sessions, tool events, project cwd,
 timestamps, and exact line-addressable source context.

 Anchor mode keeps the search grounded in the real session files. It returns sorted JSONL ranges such as:

 ```text
   count: 2
   anchors:
   - /path/to/session.jsonl:42-47 — matched all: auth; matched any: migration
 ```

 These anchors preserve the original source context and make past decisions recoverable from the line-addressable session log.

 Design

 The new mode is intentionally small and data-oriented:

 - keeps legacy behavior unchanged by default
 - uses the existing session_search tool name
 - switches behavior only through config
 - accepts a single Markdown request field
 - searches JSONL session files directly
 - filters by event timestamps, not filenames
 - supports cwd, time windows, required/alternative/excluded terms
 - returns compact path:startLine-endLine anchors with short reasons
 - does not add a persistent index or broader recall architecture

 Request format

 Anchor mode accepts only:

 ```ts
   session_search({ markdown: string })
 ```

 Example:

 ```markdown
   from: 2026-05-14
   to: 2026-05-15
   cwd: /path/to/project
   limit: 20

   all:
   - required term

   any:
   - alternative term
   - another alternative

   exclude:
   - unwanted term
 ```

 Verification

 - npm test
 - npm run check
 - npm pack --dry-run
 - smoke-tested against real Pi session JSONL files
 - verified default legacy mode remains unchanged

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ This lets Pi discover project skills as native skills without copying them into 
 
 ### Session History Search
 
-The extension indexes your Pi session history into a SQLite database with FTS5 full-text search. The agent can search across all past conversations using the `session_search` tool:
+By default, the extension indexes your Pi session history into a SQLite database with FTS5 full-text search. The agent can search across all past conversations using the `session_search` tool:
 
 | Tool | What it does |
 |---|---|---|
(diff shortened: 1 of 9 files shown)


chandra447:
This is a really nice addition — thanks for the thoughtful contribution 🙌

I really like the direction here:
- keeping the new anchor-based search **opt-in** (`sessionSearch.variant`)
- preserving existing behavior with **`legacy`** as the default
- adding solid tests around schema + anchor parsing behavior

This gives advanced users a powerful source-anchor workflow without disrupting current users. Great work, and really appreciate the contribution!

Issue #47, opened 2026-05-21
Guidelines improvements?


kdejaeger:
I'm still very noob at AI, but what I notice is that many many plugins write guidelines very 'broad', as if they are the only extension installed. And this obviously can confuse the LLM.

I think also the pi hermes guidelines can be further improved.
This is what we have atm I believe:

```
- Use the memory tool proactively when the user corrects you, shares a preference, or reveals personal details worth remembering.
- Use the memory tool when you discover environment facts, project conventions, or reusable patterns useful in future sessions.
- Do NOT use memory for temporary task state, TODO items, or session progress — only for durable, cross-session facts.
- Use target='failure' with category to save what didn't work (failures, corrections, insights).
- Use the skill tool after completing complex tasks that required trial and error or multiple tool calls.
- Use 'create' to save a new reusable procedure, 'patch' to update a section of an existing skill by skill_id, and 'update' for a full rewrite.
- Scope is required on create: choose scope='global' for transferable procedures and scope='project' when the workflow depends on this repo's paths, scripts, conventions, or deploy steps.
- Prefer structured fields for create/update: when_to_use, procedure_steps, pitfalls, and verification_steps. The tool will render valid SKILL.md sections for you.
- Use 'view' before patching or updating when you need to inspect an existing skill.
- Do NOT use skills for temporary task state — only for durable, reusable procedures.
- Use session_search when the user asks about previous discussions or past work.
- Use session_search when you need context from earlier sessions.
- Use memory_search when you need context beyond what is in the system prompt.
- Use memory_search to find project-specific memories or user preferences.
- Use memory_search with category filter to find specific types of memories (failure, correction, insight, etc.).
```

No I don't know if guidelines support indentation, I don't see that soo often. Something to research and if that is well supported we could 'group' this extensions guidlines  together maybe.

Otherwise there are some not so clear instructions that could be applicable to several extensions in the system:
```
- Use target='failure' with category to save what didn't work (failures, corrections, insights).
- Use the skill tool after completing complex tasks that required trial and error or multiple tool calls.
- Use 'create' to save a new reusable procedure, 'patch' to update a section of an existing skill by skill_id, and 'update' for a full rewrite.
- Scope is required on create: choose scope='global' for transferable procedures and scope='project' when the workflow depends on this repo's paths, scripts, conventions, or deploy steps.
[...]

PR #48, opened 2026-05-21
feat: add 5 skills to npm package with pi.skills manifest

guvengursel:
## Problem

The `pi-hermes-memory` npm package does not include its 5 document-generator skills in the distribution. The `pi.skills` manifest is missing from `package.json`, and the `skills/` directory is not listed in the `files` array.

As a result, Pi cannot auto-discover these skills from the npm installation. Users must manually install or copy skill files.

## Skills Included

- `skills/adr-recorder/SKILL.md` — Architecture Decision Records (MADR template)
- `skills/prd-writer/SKILL.md` — Product Requirements Documents
- `skills/project-architect/SKILL.md` — ARCHITECTURE.md documents
- `skills/release-prepper/SKILL.md` — Release preparation (Keep a Changelog + SemVer)
- `skills/roadmap-planner/SKILL.md` — Development roadmaps with milestones

## Changes

1. Added `pi.skills` array to `package.json` with 5 skill paths
2. Added `skills` to `package.json` `files` array
3. Added all 5 `skills/*/SKILL.md` files to the repository

## Note

These skill files already exist in the Pi local installation at `~/.pi/agent/pi-hermes-memory/skills/`. They are currently distributed via a separate install path, not through npm. This PR makes them available directly from the npm package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "src",
     "README.md",
     "LICENSE",
-    "docs"
+    "docs",
+    "skills"
   ],
   "pi": {
     "extensions": [
@@ -58,5 +59,12 @@
   "dependencies": {
     "@earendil-works/pi-tui": "^0.74.0",
     "better-sqlite3": "^12.9.0"
-  }
+  },
+  "pi.skills": [
+    "skills/adr-recorder/SKILL.md",
+    "skills/prd-writer/SKILL.md",
+    "skills/project-architect/SKILL.md",
+    "skills/release-prepper/SKILL.md",
+    "skills/roadmap-planner/SKILL.md"
+  ]
 }
--- a/skills/adr-recorder/SKILL.md
+++ b/skills/adr-recorder/SKILL.md
@@ -1,0 +1,217 @@
+---
+name: adr-recorder
+description: Write Architecture Decision Records (ADRs) using the MADR template. Use when the user makes a significant technical decision (framework choice, database, deployment platform, architecture pattern, library selection) and wants it documented for future reference. ADRs answer "why did we do this?" months later.
+metadata:
+  version: 2
+  created: "2026-05-19"
+  updated: "2026-05-19"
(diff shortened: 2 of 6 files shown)


kdejaeger:
This also maybe mean we don't need that much bloat in the system prompt? See #47

chandra447:
@guvengursel — thank you for this, and I'm sorry it sat open this long without a real answer. That's on me, and you deserved a decision months ago rather than silence.

I'm closing it, and I want to give you the actual reasoning rather than a stale-bot brush-off.

**1. The mechanism works — I verified it.** `pi.skills` is genuinely supported: `readPiManifest()` reads the `pi` key from a package's `package.json`, and `collectPackageResources()` expands each `skills` entry, falling back to a conventional `skills/` directory when no manifest exists. Your PR was technically correct.

**2. But it adds a third skill source, right as we're removing the second.** Issues #125 and #126 turned out to be a real bug, not just warning noise. Pi keys skills **by name**, first-loaded wins, and drops the loser. Our global skills lived in a private directory that lost every collision against `~/.pi/agent/skills/` — so `skill_manage` edits on a shadowed name silently did nothing. #131 collapses us down to exactly one global skills location. Shipping five package-level skills would reintroduce a separate source with its own five names competing in the same map.

**3. And it pushes the wrong way on prompt budget.** @kdejaeger flagged this in #47 and I think they're right — this is a *memory* extension whose value proposition is token discipline, and it already carries the largest guidelines block of any extension a user is likely to install. Five always-loaded document-generator skills, none of which are about memory, is the opposite of that.

**4. The skills themselves aren't the problem.** ADR recorder, PRD writer, project architect, release prepper, roadmap planner — those are legitimately useful, they're just not *memory* skills. They'd be better as their own package that a user opts into. If you publish them that way, I'll link it from our README, and I mean that.

Your PR is also what made me go read Pi's package-resource code properly, which is directly why the #125/#126 fix is grounded in the real implementation instead of guesswork. That was worth having.

PR #49, opened 2026-05-21
docs(memory-search): document FTS5 query syntax

Istar-Eldritch:
## Problem

The `query` parameter of `memory_search` is passed directly to SQLite FTS5, but this was not documented. Agents passing multi-word natural language queries (e.g. `database migration`) expect fuzzy/AND matching, but FTS5 treats unquoted multi-word input as a phrase search — requiring the words to appear consecutively. This causes searches to silently return zero results even when relevant entries exist.

## Fix

Document the FTS5 query syntax in both the tool description and the parameter description so agents construct effective queries:

- Single word: `auth`
- Exact phrase: `"database migration"`
- Proximity: `NEAR("deploy" "config", 10)`
- Boolean: `auth AND token`, `auth OR oauth`, `auth NOT oauth`

No code change — the existing FTS5 passthrough already supports all of this.

### DIFF
--- a/src/tools/memory-search-tool.ts
+++ b/src/tools/memory-search-tool.ts
@@ -24,15 +24,24 @@ Use cases:
 - Find user preferences: "What are the user's testing preferences?"
 - Search for past failures: "memory_search('auth', category='failure')"
 
+Query syntax (FTS5):
+- Single word: auth
+- Exact phrase: "database migration" (quotes required for phrase matching)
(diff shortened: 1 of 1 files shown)


Istar-Eldritch:
Closing in favour of an issue — the docs-only approach is insufficient since agents can't reliably infer FTS5 syntax from the description alone. The underlying escapeFts5Query fix (quoting individual terms instead of the whole phrase) is the correct solution.

PR #51, opened 2026-05-23
Fix project memory auto-consolidation

ffrappo:
## Summary
- wire auto-consolidation into the project-scoped MemoryStore
- tell child consolidation runs to use target='project' for project memory
- include project memory in /memory-consolidate when a project store is available
- add coverage for project consolidation prompts and manual command behavior

## Why
Project memory writes route to projectStore, but only the global store had a consolidator. When project memory hit the character limit, auto-consolidate was skipped and the memory tool returned the capacity error.

## Test
- npm run check -- --pretty false
- node --import tsx --test tests/handlers/auto-consolidate.test.ts
- npm test

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -12,24 +12,38 @@ import { MemoryStore } from "../store/memory-store.js";
 import { CONSOLIDATION_PROMPT, ENTRY_DELIMITER } from "../constants.js";
 import type { ConsolidationResult } from "../types.js";
 
+type MemoryTarget = "memory" | "user" | "failure";
+type ToolMemoryTarget = MemoryTarget | "project";
+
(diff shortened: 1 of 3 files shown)


chandra447 (approved):

chandra447:
Thank you for contributing this and for actively using the extension — really appreciated. 🙏



Strengths:
- Adds project-aware consolidation target routing (target = "project")
- Wires consolidator for projectStore in src/index.ts
- Extends /memory-consolidate to include project scope output
- Adds targeted tests for project prompt labeling/targeting and command integration

Non-blocking nits: (can you please open a follow-up PR )
1) entriesForTarget() currently maps non-user targets to getMemoryEntries(). If failure is ever routed here, failure entries are not read explicitly. Either narrow supported target values or add explicit failure handling + a test.
2) Optional regression test: validate /memory-consolidate output when projectStore is absent, to lock legacy behavior.

PR #57, opened 2026-05-27
fix: quote individual FTS5 terms instead of entire query (fixes #50)

### LINKED ISSUE #50: memory_search returns zero results for multi-word queries
## Bug

`memory_search` silently returns zero results for any multi-word natural language query, even when the database contains hundreds of relevant entries.

## Root Cause

`escapeFts5Query` in `src/store/sqlite-memory-store.ts` wraps the entire query in double quotes:

```ts
return `"${query.replace(/"/g, '""')}"`;
```

This makes every query an **exact FTS5 phrase search**, requiring all words to appear consecutively in the stored text. Queries like `"linear cli"` or `"memory search"` almost never match because memory entries don't contain those exact word sequences.

## Reproduction

With 161 entries in the DB:
- `memory_search("linear")` → 18 results ✓
- `memory_search("linear cli")` → 0 results ✗ (phrase search fails)
- `memory_search("memory search")` → 0 results ✗

## Proposed Fix

Quote individual terms instead of the whole query, so FTS5 uses implicit AND (all terms must appear, in any order):

```ts
const terms = query.trim().split(/\s+/).filter(Boolean);
if (terms.length === 0) return '""';
return terms.map(t => `"${t.replace(/"/g, '""')}"`).join(' ');
```

This preserves single-word behaviour and FTS5 operator pass-through (OR, AND, NOT, NEAR) while fixing the common multi-word case. Exact phrase matching is still available to callers who explicitly wrap their query in quotes.

atimofeev:
## Problem

`escapeFts5Query()` in both `sqlite-memory-store.ts` and `session-search.ts` wrapped the entire query in double quotes, making FTS5 require an **exact phrase match**. Multi-word natural language queries like `gpu issues` or `laptop hardware` almost never produce results because the exact phrase doesn't appear in stored text.

## Reproduction

With 400+ indexed messages:
- `memory_search("nvidia")` → 29 results ✓
- `memory_search("gpu issues")` → **0 results** ✗ (phrase match fails)
- `session_search("laptop hardware")` → **0 results** ✗

## Fix

Quote **individual terms** so FTS5 uses implicit AND matching (all terms must appear, in any order):

```ts
// Before:
return `"${query.replace(/"/g, '""')}"`;

// After:
const terms = query.trim().split(/\s+/).filter(Boolean);
if (terms.length === 0) return '""';
return terms.map(t => `"${t.replace(/"/g, '""')}"`).join(' ');
```

Also added `query.includes('"')` check to the operator passthrough condition — callers who wrap their query in explicit double quotes still get exact phrase matching.

## Behavior

| Query | Before | After |
|-------|--------|-------|
| `gpu` | matches | matches (unchanged) |
| `gpu issues` | `"gpu issues"` → 0 hits | `"gpu" "issues"` → AND match |
| `auth OR token` | passthrough | passthrough (unchanged) |
| `"nvidia modeset"` | passthrough | passthrough (new) |

## Files

- `src/store/sqlite-memory-store.ts` — `memory_search` tool
- `src/store/session-search.ts` — `session_search` tool

Closes #50

### DIFF
--- a/src/store/session-search.ts
+++ b/src/store/session-search.ts
@@ -2,15 +2,19 @@ import { DatabaseManager } from './db.js';
 
 /**
  * Escape a string for FTS5 query syntax.
- * Wraps the query in double quotes to treat it as a literal phrase.
+ * Quotes individual terms so FTS5 uses implicit AND (all terms must appear, any order).
+ * Preserves explicit FTS5 operators (OR, AND, NOT, NEAR) and quoted phrases.
  */
 function escapeFts5Query(query: string): string {
-  // If the query already contains FTS5 operators (OR, AND, NOT, NEAR), leave it as-is
-  if (/\b(OR|AND|NOT|NEAR)\b/.test(query)) {
+  // If the query already contains FTS5 operators (OR, AND, NOT, NEAR) or explicit
+  // double-quoted phrases, pass through as-is so callers retain full control.
+  if (/\b(OR|AND|NOT|NEAR)\b/.test(query) || query.includes('"')) {
     return query;
   }
-  // Otherwise, wrap in double quotes to treat as literal phrase
-  return `"${query.replace(/"/g, '""')}"`;
+  // Quote individual terms so FTS5 matches all of them with implicit AND.
+  const terms = query.trim().split(/\s+/).filter(Boolean);
+  if (terms.length === 0) return '""';
+  return terms.map(t => `"${t.replace(/"/g, '""')}"`).join(' ');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


chandra447:
Thanks for the fix here — the multi-word FTS5 issue is real, and the implicit-AND direction looks right.

I do see one regression risk before merge:

- `query.includes("\"")` is too broad as a passthrough condition. In `memory_search`, the query is forwarded straight into `searchMemories()` with no FTS error handling, so a natural-language query with an unmatched quote like `issue "timeout` will now raise an FTS parse error instead of behaving like a normal search. `session_search` is insulated because it catches malformed FTS queries, but `memory_search` is not.

I think this should either:
- only passthrough when the quotes are balanced / clearly intentional phrase syntax, or
- keep escaping by default and add a narrower explicit-phrase condition.

It would also help to add regression coverage for:
- multi-word implicit-AND matching
- explicit quoted phrase matching
- malformed quote input

PR #65, opened 2026-06-04
fix: broaden memory_search fallback for natural-language queries

### LINKED ISSUE #64: The tool to retrieve memories conflicts with builtin pi memory_search
As per the object of the issue...this extension seems to work flawlessly while saving memories and updating them...but fails to retrieve them with simple queries ("Who am I"). I think, mostly is Qwen3.6 detecting it, that there's a collision between the builtin tool and the tool exposed by this extension

The builtin memory search never queries the sqlite db created from pi-hermes-memory 

Image

chandra447:
## Summary
- add a natural-language fallback query for memory search when strict FTS matching returns no results
- preserve explicit FTS operator queries without broadening them
- add regression coverage for the Naruto/identity-style retrieval miss and bump package version to 0.7.15

## Testing
- npm test
- npm run check

Closes #64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ Thumbs.db
 docs/0.4/RELEASE-POST.md
 docs/0.4/linkedin_post.md
 .codex
+.serena
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pi-hermes-memory",
-  "version": "0.7.14",
+  "version": "0.7.15",
   "lockfileVersion": 3,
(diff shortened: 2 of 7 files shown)


chandra447:
Fixed and released in `v0.7.15`.

Thanks again for raising and reproducing this clearly — the screenshot made it much easier to separate the real retrieval problem from the initial tool-collision suspicion.

The fix keeps the strict SQLite FTS query first, then retries with a broader natural-language fallback only when that strict pass returns no results. Explicit operator queries still behave exactly as written.

PR #67, opened 2026-06-10
fix: honor PI_CODING_AGENT_DIR for agent paths

nekwebdev:
## Summary

- Route Hermes default storage paths through a shared agent-root resolver.
- Prefer `PI_CODING_AGENT_DIR` when present, with `~/...` expansion and legacy `~/.pi/agent` fallback.
- Update config, session indexing/search, memory store, skill store, project switching, and related tests to use the resolved agent root.

## Motivation

Hermes still hard-coded several defaults under `~/.pi/agent`. That breaks isolated Pi agent environments that set `PI_CODING_AGENT_DIR`, causing memory, skills, sessions, and config files to land outside the active agent directory.

This change centralizes agent-root resolution so all default paths follow the configured Pi agent environment while preserving backward compatibility for existing users.

## Changes

- Add `resolveAgentRoot()` in `src/paths.ts`.
- Export `AGENT_ROOT` from that resolver.
- Use `AGENT_ROOT` for default config, session, memory, skill, and project-memory paths.
- Use `resolveProjectsRoot()` in project switching instead of rebuilding paths manually.
- Add path-resolution tests for:
  - `PI_CODING_AGENT_DIR` override
  - blank/unset fallback
  - home-relative path expansion
- Update existing tests to assert against resolved agent-root paths instead of hard-coded `~/.pi/agent` strings.

## Testing

- `npm test` — all 33 test files passed.
- `npm run check` — TypeScript check passed.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,5 @@
 import * as fs from "node:fs";
 import * as path from "node:path";
-import * as os from "node:os";
 import type { MemoryConfig, MemoryOverflowStrategy, SessionSearchVariant, ThinkingLevel } from "./types.js";
 import {
   DEFAULT_MEMORY_CHAR_LIMIT,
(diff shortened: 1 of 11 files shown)


chandra447 (approved):
Verified locally.
- Confirmed agent-root resolution now consistently honors PI_CODING_AGENT_DIR with fallback to ~/.pi/agent\n- Confirmed hardcoded agent-root paths were replaced across config, session indexing, memory/skill stores, and session search
- Ran: npm run check plus targeted tests for paths, config, project detection, resource discovery, and extension-root migration
 
- This looks like a safe non-breaking compatibility fix that makes custom Pi agent directories behave correctly.

PR #75, opened 2026-06-12
fix(db): close SQLite on session shutdown to truncate the WAL

### LINKED ISSUE #74: WAL file is never truncated: session_shutdown does not close the SQLite connection
## Summary

The SQLite WAL file (`sessions.db-wal`) is never truncated, so it grows to its high-water mark and is never reclaimed across sessions. In one environment it sat at ~4.3 MB next to a 6.1 MB `sessions.db`.

## Root cause

`DatabaseManager.close()` runs `PRAGMA wal_checkpoint(TRUNCATE)`, but `close()` is never called. The only `session_shutdown` handler that touches the DB (`src/index.ts`, "Auto-index session on shutdown") indexes the session and returns without closing the connection:

```ts
pi.on("session_shutdown", async (_event, ctx) => {
  try {
    // ... indexSession(dbManager, sessionData)
  } catch {
    // Silent fail
  }
});
```

`PRAGMA wal_autocheckpoint = 100` (added in #56) passively checkpoints pages back into the main DB, but it does not truncate the `-wal` file — that only happens on a `TRUNCATE` checkpoint or a clean connection close. Since the process exits with the connection still open, the WAL file persists at its largest size. This complements #56 rather than duplicating it.

## Fix

Close the connection in a `finally` block of the last-registered `session_shutdown` handler, so it runs after the session-flush handler's writes and is the final DB activity. `close()` already issues the `TRUNCATE` checkpoint.

## Impact

- Stale WAL storage that never shrinks.
- A larger WAL replay window on next open.

A PR with the fix and a regression test (asserts the WAL is 0 bytes after `close()`) follows.

jzhuge:
Fixes #74.

## Problem

`DatabaseManager.close()` runs `PRAGMA wal_checkpoint(TRUNCATE)`, but `close()` is never called. The `session_shutdown` auto-index handler indexes the session and returns without closing the connection, so the `-wal` file grows to its high-water mark and is never reclaimed across sessions (observed ~4.3 MB next to a 6.1 MB `sessions.db`). The `wal_autocheckpoint` added in #56 passively checkpoints pages but does not truncate the WAL file; only a `TRUNCATE` checkpoint / clean close does.

## Fix

Close the DB connection in a `finally` block of the auto-index `session_shutdown` handler. It is registered last, so it runs after the session-flush handler's writes and is the final DB activity — the correct point to checkpoint. `close()` already issues the `TRUNCATE` checkpoint and is safe to call repeatedly.

The ordering is guaranteed, not assumed: Pi's `ExtensionRunner.emit()` runs same-extension `session_shutdown` handlers sequentially in registration order and `await`s each one (no `Promise.all`), so the flush writes fully complete before `close()` runs. A maintainer warning in the comment notes not to register a DB-writing `session_shutdown` handler after the close block.

## Test

Adds `tests/store/db.test.ts` coverage: writes enough rows to grow a non-trivial WAL, then asserts the `-wal` file is 0 bytes after `close()`. The WAL path is derived from `dbManager.getPath()` rather than hardcoded.

## Verification

- `npm run check` (tsc) — clean
- full suite via `tests/run-all.sh` — 32/32 files pass, including the new test

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -211,6 +211,12 @@ export default function (pi: ExtensionAPI) {
   // is the final DB activity. Closing here truncates the WAL via
   // PRAGMA wal_checkpoint(TRUNCATE); without it the WAL only grows to its
   // high-water mark and is never reclaimed across sessions.
+  //
+  // Ordering is safe: Pi's ExtensionRunner.emit() runs same-extension handlers
+  // sequentially in registration order and awaits each one, so the flush above
+  // fully completes before close() runs. WARNING: do not register another
+  // DB-writing session_shutdown handler after this block — it would run after
+  // close() and silently no-op.
   pi.on("session_shutdown", async (_event, ctx) => {
     try {
       const sessionFile = ctx.sessionManager.getSessionFile();
--- a/src/index.ts
+++ b/src/index.ts
@@ -207,6 +207,10 @@ export default function (pi: ExtensionAPI) {
   registerIndexSessionsCommand(pi);
 
   // ── 11. Auto-index session on shutdown ──
+  // Registered last, so this runs after the session-flush shutdown handler and
+  // is the final DB activity. Closing here truncates the WAL via
+  // PRAGMA wal_checkpoint(TRUNCATE); without it the WAL only grows to its
+  // high-water mark and is never reclaimed across sessions.
   pi.on("session_shutdown", async (_event, ctx) => {
     try {
       const sessionFile = ctx.sessionManager.getSessionFile();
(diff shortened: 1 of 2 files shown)


chandra447 (approved):
Confirmed the shutdown handler now closes the DatabaseManager in a finally block after session flush work completes
- Confirmed the new regression test proves the WAL is truncated to 0 bytes on close
- Ran: npm run check && npx tsx --test tests/store/db.test.ts
 
- This looks like a safe non-breaking reliability fix that keeps WAL files from lingering across sessions.

PR #78, opened 2026-06-17
perf: only load pi-hermes-memory extension in child subprocesses

xuli500177:
## Problem

The background review, correction detection, and session flush subprocesses (`pi -p --no-session`) currently load **all** extensions from `settings.json` packages — including context-mode, pi-lens, pi-web-access, pi-review, pi-cache-optimizer, etc. This adds ~20K+ tokens of system prompt overhead and significant startup CPU for a simple one-shot "review and decide if memory needs saving" LLM call.

This is especially wasteful for users with many plugins installed.

## Solution

Pi already has a `--no-extensions` (`-ne`) CLI flag that skips loading all packages from `settings.json`, keeping only extensions explicitly passed via `-e` / `--extension`.

This PR:
1. Resolves pi-hermes-memory's own extension entry path via `import.meta.url`
2. Passes `--no-extensions` to the child subprocess to skip all settings.json packages
3. Passes `-e <own_path>` so only pi-hermes-memory is loaded
4. Model/thinking overrides (llmModelOverride, llmThinkingOverride) are unchanged — they still work as before

## Impact

| Before | After |
|---|---|
| Loads ALL plugins (~15-20 extensions) | Loads ONLY pi-hermes-memory (~1 extension) |
| ~25K+ tokens system prompt | ~3-5K tokens system prompt |
| Slow startup (LSP init, MCP servers, etc.) | Fast startup (just memory tool) |

## Testing

All 10 existing tests pass with updated assertions.

Closes #(add issue number)

### DIFF
--- a/src/handlers/pi-child-process.ts
+++ b/src/handlers/pi-child-process.ts
@@ -1,5 +1,5 @@
 import { existsSync } from "node:fs";
-import { dirname, join } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import type { MemoryConfig, ThinkingLevel } from "../types.js";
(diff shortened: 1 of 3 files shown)


chandra447:
Rebased this PR onto the latest `main` after #81 and #82 landed.

What changed during the rebase:

- Resolved conflicts in `src/handlers/pi-child-process.ts` and related tests.
- Preserved the Windows child Pi invocation fix from #80 (`node.exe` + resolved `cli.js` path on Windows).
- Kept this PR’s intended behavior: child subprocesses now run with `--no-extensions -e <pi-hermes-memory entrypoint>` so they load only this extension instead of all configured Pi extensions.
- Ensured the retry path also uses the minimal extension args while still dropping model/thinking overrides.
- Updated a stale auto-consolidation test assertion to verify `--model` / `--thinking` are removed on retry without assuming a fixed arg length.

Validation run after rebase:

- `node --import tsx --test tests/handlers/pi-child-process.test.ts` ✅
- `node --import tsx --test tests/handlers/background-review.test.ts` ✅
- `node --import tsx --test tests/handlers/auto-consolidate.test.ts` ✅
- `npm run check` ✅
- `npm test` ✅ all 35 test files passed
- `git diff --check` ✅

The branch has been force-pushed to the PR head and now applies cleanly on top of latest `main`.

PR #84, opened 2026-06-22
fix(session-search): bound startup backfill

### LINKED ISSUE #83: Startup session backfill blocks/slows Pi with large session history
## Summary

`pi-hermes-memory@0.7.18` runs a startup session backfill from the `session_start` hook. With a large `~/.pi/agent/sessions` directory this can cause a major Pi startup / early-session delay.

The visible notification is:

```text
🧠 Session backfill complete: 0 indexed, 824 skipped, 0 messages.
```

Even when all sessions are skipped, the extension still scans/parses many JSONL files.

## Environment

- Package: `pi-hermes-memory`
- Version: `0.7.18`
- Pi install: global package via `settings.json`
- Session history size in one affected environment:
  - `1783` JSONL session files
  - about `6.9 GiB` under `~/.pi/agent/sessions`
  - rough local parse pass took about `30s`

## Observed behavior

On Pi startup, Hermes schedules session backfill here:

```ts
// src/index.ts
pi.on("session_start", async (_event, ctx) => {
  ...
  scheduleSessionBackfill(dbManager, sessionsDir, ...)
});
```

The scheduler says the work is non-blocking because it uses `setTimeout(0)`:

```ts
// src/handlers/session-backfill.ts
setTimeoutFn(() => {
  const result = indexAllSessionsFn(dbManager, sessionsDir);
  ...
}, 0);
```

But `indexAllSessions()` is synchronous and scans/parses all session files:

```ts
// src/store/session-indexer.ts
for (const file of files) {
  const session = parseSessionFile(file);
  const indexResult = indexSession(dbManager, session);
  ...
}
```

So `session_start` may return, but the process immediately performs heavy synchronous disk/JSON work. In the TUI this still feels like startup or early chat is blocked.

## Expected behavior

Pi startup should not synchronously scan and parse the full session history.

Previously the v0.4 design docs explicitly said:

```text
Don't parse JSONL files on every startup
Index a session when session_shutdown fires
Bulk import existing sessions via /memory-index-sessions command
```

That expectation still seems right for large session histories.

## Preferred fix

Replace full startup parsing with bounded incremental backfill.

At startup, Hermes should avoid parsing every JSONL session file. It can do a cheap discovery pass using filesystem metadata and only parse files that are likely missing or changed.

Possible design:

1. Add lightweight file metadata to the session index, for example:
   - session file path
   - file size
   - file mtime
   - last indexed timestamp

2. During startup backfill:
   - list session JSONL files
   - compare each file's current `size` / `mtime` with stored metadata
   - skip files whose metadata has not changed
   - parse only files that are new or changed

3. Cap startup work:
   - process at most N files per startup, e.g. 25 or 50
   - optionally cap elapsed time, e.g. 500–1000ms
   - continue remaining work on later startups or via `/memory-index-sessions`

4. Keep `/memory-index-sessions` as the explicit full backfill command.

This keeps automatic indexing useful for missed/crashed sessions while preventing startup from synchronously parsing a large historical archive.

## Acceptance criteria

- Startup backfill does not parse unchanged session JSONL files.
- Startup backfill has a bounded file count or time budget.
- New or changed session files are still indexed automatically.
- `/memory-index-sessions` still performs a full manual backfill.
- Existing live/current session indexing still works.
- Large archives with thousands of JSONL files do not cause multi-second startup stalls.

AndriiSemenkov:
## Summary

Fixes #83 by changing startup session backfill from a full JSONL archive parse into a bounded incremental pass.

## What changed

- Add a `session_files` metadata table with path, session id, size, mtime, and indexed timestamp.
- Record file metadata during full/manual session indexing and live session indexing.
- Add `indexChangedSessions()` for startup backfill:
  - skips unchanged files by stored size/mtime
  - seeds metadata for already-indexed sessions from Pi JSONL filenames without parsing the body
  - parses only new/changed files
  - caps parsed files per startup at 50
- Keep `/memory-index-sessions` using full `indexAllSessions()` for explicit manual backfill.
- Avoid marking startup backfill complete when the cap is reached, so future startups can continue.
- Document the bounded incremental startup backfill behavior.

## Validation

- [x] `npm run check`
- [x] `npm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ Search behavior notes:
 - Exact phrases can be requested with quotes, for example `"memory search"`.
 - Advanced FTS queries with operators like `OR` still work when you need them.
 
-Session history is indexed automatically during the active session and on session shutdown. Startup also runs a bounded incremental backfill for missed sessions: it compares stored file metadata and only parses new or changed JSONL files, capped per startup. To bulk-import existing sessions manually:
+Session history is indexed automatically during the active session and on session shutdown. Startup also runs a bounded incremental backfill for missed sessions: it compares stored file metadata and only parses files without matching metadata, capped per startup. To bulk-import existing sessions manually:
 
 ```
 /memory-index-sessions
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ Search behavior notes:
 - Exact phrases can be requested with quotes, for example `"memory search"`.
 - Advanced FTS queries with operators like `OR` still work when you need them.
 
-Session history is indexed automatically on session shutdown. To bulk-import existing sessions:
+Session history is indexed automatically during the active session and on session shutdown. Startup also runs a bounded incremental backfill for missed sessions: it compares stored file metadata and only parses new or changed JSONL files, capped per startup. To bulk-import existing sessions manually:
 
 ```
 /memory-index-sessions
--- a/src/handlers/session-backfill.ts
+++ b/src/handlers/session-backfill.ts
@@ -1,12 +1,13 @@
 import type { DatabaseManager } from '../store/db.js';
 import {
-  indexAllSessions,
+  indexChangedSessions,
   needsBackfill,
   touchBackfillTimestamp,
   type BulkIndexResult,
 } from '../store/session-indexer.js';
 
 export const SESSION_BACKFILL_SHUTDOWN_TIMEOUT_MS = 5000;
+export const SESSION_BACKFILL_MAX_FILES = 50;
 
 type NotifyLevel = 'info' | 'warning' | 'error';
 type NotifyFn = (message: string, level: NotifyLevel) => void;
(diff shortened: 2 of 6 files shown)


AndriiSemenkov:
Manual verification against a temp `PI_CODING_AGENT_DIR` with `sessions` symlinked to my real session archive and a copied `sessions.db` with `last_session_backfill` removed:

### Direct backfill benchmark

- Fixed `needsBackfill`: `13ms`
- Fixed `indexChangedSessions({ maxFilesToIndex: 50 })`: `47ms`
  - result: `0 indexed, 825 skipped, 0 messages`
- Old `indexAllSessions`: `19865ms`
  - result: `0 indexed, 825 skipped, 0 messages`

### Pi CLI smoke

Used `--no-extensions -e <hermes> -e /tmp/pi-exit-after-start.ts --offline --no-session` with a tiny extension that exits 100ms after `session_start`. Because Hermes schedules backfill with `setTimeout(0)`, the old full scan blocks that exit timer.

- Local PR extension: `real 0.82s`
- Installed `pi-hermes-memory@0.7.18`: `real 21.20s`

This confirms the PR avoids the startup/early-session full JSONL parse while preserving the backfill path.

AndriiSemenkov:
Additional manual verification for the "one genuinely new session file" case:

Setup: temp `sessions` dir with symlinks to my real session archive plus one synthetic new Pi JSONL session file not present in the copied DB.

Results:

- Fixed `needsBackfill`: `44ms`, returned `true`
- Fixed `indexChangedSessions({ maxFilesToIndex: 50 })`: `51ms`
  - result: `1 processed, 1 indexed, 825 skipped, 1 message`
  - verified new message row exists in SQLite: `count = 1`
- Old `indexAllSessions`: `19578ms`
  - result: `826 processed, 1 indexed, 825 skipped, 1 message`
  - verified new message row exists in SQLite: `count = 1`

So the incremental path indexes the new conversation while avoiding the historical JSONL parse; the old path indexes the same new conversation only after parsing the whole archive.

chandra447:
Reviewing against issue #83 — this correctly fixes the root cause. Startup now does stat-only discovery (`needsBackfill` stats files, no JSONL parse) + bounded parse (`indexChangedSessions`, cap 50) instead of `indexAllSessions`, and `/memory-index-sessions` retains the full backfill. The `reachedLimit` → skip `touchBackfillTimestamp` logic correctly resumes across startups. Benchmarks check out.

Merging this as the core fix. Filing a follow-up PR for two gaps identified during review:

**Gap 1 — No newest-first ordering.** `indexChangedSessions` iterates files in filesystem order and breaks at the cap. A recently-crashed session at the end of the listing could wait dozens of startups to be indexed while the cap is consumed by old historical files at the front. Crash recovery is the scenario where startup backfill adds the most value (the live `message_end` path missed the final state), so this matters. Fix: sort the changed set by `mtimeMs` desc before applying the cap. Crashes are recent → indexed on the next startup; old catch-up trickles in after.

**Gap 2 — `session_shutdown` handler bypasses metadata upsert.** `src/index.ts` shutdown handler calls `indexSession(...)` directly, which does not upsert `session_files` metadata (only `indexLiveSession` / `indexSessionFile` do). Pi appends the final session entry on shutdown after the last `message_end`, so the file size/mtime changes but `session_files` keeps the stale live-index value. Next startup sees a mismatch → re-parses every recently-closed session. Fix: route through `indexLiveSession` or call `upsertSessionFileMetadata` explicitly (needs exporting).

Neither blocks the core fix, so merging now and addressing both in a follow-up.

Issue #86, opened 2026-06-23
Live session indexing warns forever once sessions.db is corrupt — open() has no integrity check / self-heal


boeykokchin1969:
## Summary

Once the shared `sessions.db` becomes corrupt, **every** `message_end` re-runs the live indexer, hits the malformed pages, and logs:

```
⚠️ Live session indexing failed: database disk image is malformed
```

…on **every turn, indefinitely**, with no recovery. The warning is non-fatal, but it never stops until the DB is manually repaired/recreated, and live indexing is silently broken for the whole period. Root cause: `DatabaseManager.open()` (`src/store/db.ts`) opens and migrates the DB but **never checks integrity and has no `SQLITE_CORRUPT` handling**, so a torn DB stays torn.

## Environment

- `pi-hermes-memory` 0.7.19, Pi coding agent, `better-sqlite3`
- Single shared DB: `<AGENT_ROOT>/pi-hermes-memory/sessions.db`, `journal_mode=WAL`, `wal_autocheckpoint=100`
- Host: WSL2 (Linux ext4 on a Windows `.vhdx`)

## What happens

- `src/index.ts` registers `pi.on("message_end", … scheduleLiveSessionIndex(dbManager, …, { onError: (err) => console.warn(\`⚠️ Live session indexing failed: ${err.message}\`) }))`.
- `indexLiveSession` (`src/store/session-indexer.ts`) does `dbManager.getDb()` → prepared `INSERT`s inside `db.transaction(...)`.
- If `sessions.db` is corrupt, the inserts throw `SQLITE_CORRUPT: database disk image is malformed`, caught by `onError` → the warning. Next `message_end` repeats it. Forever.

## Root cause

`DatabaseManager.open()` runs `PRAGMA journal_mode=WAL` + schema/migrations, then returns the handle. There is **no `PRAGMA quick_check`/`integrity_check`** and **no `SQLITE_CORRUPT` catch** anywhere in the DB layer. So a DB that was torn by an earlier crash is opened "successfully" (SQLite opens structurally-damaged DBs without throwing on open), and only fails — repeatedly — on subsequent writes.

## How the DB gets corrupt (context)

This is a **torn write** from abrupt termination (WSL2/host shutdown, power loss, hibernate, OOM/SIGKILL) interrupting a write or WAL-checkpoint. The corruption signature on an affected `sessions.db` is pure b-tree pointer damage — **not** FTS5, **not** a logic bug:

```
*** in database main ***
[...]

PR #88, opened 2026-06-25
chore: sync package-lock version to 0.7.20

chandra447:
Follow-up to #86 (commit 3d82f44).

`package.json` was bumped to `0.7.20` when the DB corruption self-heal fix landed, but `package-lock.json` was left at `0.7.19`. This syncs both root `version` fields in the lockfile so they stay consistent.

No code changes — lockfile-only.


chandra447:
Superseded by #87 — the lockfile version bump is included in that PR's commit set.

PR #97, opened 2026-07-07
fix: lock memory consolidation subprocesses

paxcalpt:
## Summary
- add an atomic per-target lock around auto-consolidation subprocesses
- skip duplicate child Pi launches when another consolidation is already running for the same target
- clean stale lock directories after the configured timeout plus a grace period
- add a regression test that verifies concurrent consolidation only spawns one child process

## Verification
- npm run check
- node --test --import tsx tests/handlers/auto-consolidate.test.ts
- npm test

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -7,14 +7,84 @@
  * from disk after consolidation completes.
  */
 
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import { MemoryStore } from "../store/memory-store.js";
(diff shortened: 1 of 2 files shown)


robinbraemer:
I incorporated this exact contributor commit (preserving Ricardo's Git authorship) into the stacked hardening PR #102, then layered the confirmed lock-identity, ownership, release, and Copilot-review fixes there. Please coordinate follow-up work on #102 so we do not create duplicate fixing lanes.

chandra447:
Landed via #102 ("fix: serialize memory mutations and consolidation"), which explicitly preserved this PR's original commit and authorship while hardening it further against races found during review. Thanks for the fix.

chandra447:
Thanks for raising this and for the underlying fix — your commit was incorporated (with authorship preserved) into the stacked hardening in #102, which merged. Appreciate you finding this and being open to folding it into the combined fix rather than duplicating effort.

PR #99, opened 2026-07-08
fix: detect external file writes before memory mutations (consolidation race)

overpod:
## Summary

`MemoryStore.add/replace/remove` mutate stale in-memory arrays and save them to disk without checking whether the markdown file was changed by an external writer — a consolidation child process, another pi session, or a manual edit. 

**Real-world impact (reproduced on v0.7.23):** while `/memory-consolidate` is still consolidating one target (child pi processes run up to 180s each), a `memory add` in the live session reports success but the entry is **silently lost** when the file is rewritten from a stale snapshot; conversely, a stale in-session save can clobber freshly consolidated content.

## Fix

- Track a per-file fingerprint (`mtimeMs:size`) at every `loadFromDisk`/`saveToDisk`
- New `syncTargetFromDiskIfChanged()` runs before every mutation (`_add`/`replace`/`remove`): if the fingerprint no longer matches, re-read just that target from disk so the mutation applies on top of the current on-disk state
- The frozen system-prompt snapshot is intentionally **not** rebuilt on re-sync (prompt-cache safety)
- Consolidation children run the same store code, so the race is closed on both sides

Complementary to #97 — that PR prevents duplicate consolidation subprocesses per target; this one protects live-session mutations against any external file writer.

## Verification

- `npm run check` — clean
- `npm test` — all 36 test files pass
- 4 new regression tests in `tests/store/memory-store.test.ts` (`external file changes (consolidation race)`): on unpatched `memory-store.ts` the 3 race tests fail, with the patch all pass

## Repro (before the fix)

1. Fill memory near the limit, run `/memory-consolidate`
2. While a slower target (e.g. `failure`) is still consolidating, call `memory add` — returns `success`
3. Entry is gone from MEMORY.md after the consolidation child's write / next `loadFromDisk`

### DIFF
--- a/src/store/memory-store.ts
+++ b/src/store/memory-store.ts
@@ -32,6 +32,8 @@ export class MemoryStore {
   private userEntries: string[] = [];
   private failureEntries: string[] = [];
   private snapshot: MemorySnapshot = { memory: "", user: "" };
+  /** Per-file fingerprint (mtimeMs:size) captured at last load/save — detects external writes. */
+  private fileFingerprints: Record<string, string> = {};
   private consolidator: ((target: "memory" | "user" | "failure", signal?: AbortSignal) => Promise<ConsolidationResult>) | null = null;
 
   constructor(private config: MemoryConfig) {}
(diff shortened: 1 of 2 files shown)


robinbraemer:
I incorporated this exact contributor commit (preserving Trofimov Alex's Git authorship) into the stacked hardening PR #102, then layered the confirmed publish-boundary, rollback, alias, and recovery fixes there. Please coordinate follow-up work on #102 so we do not create duplicate fixing lanes.

chandra447:
Landed via #102 ("fix: serialize memory mutations and consolidation"), which explicitly preserved this PR's original commit and authorship while hardening it further against races found during review. Thanks for the fix.

chandra447:
Thanks for raising this and for the underlying fix — your commit was incorporated (with authorship preserved) into the stacked hardening in #102, which merged. Appreciate you finding this and being open to folding it into the combined fix rather than duplicating effort.

PR #100, opened 2026-07-11
fix: bound session search output

robinbraemer:
## Summary

- bound each legacy `session_search` snippet to 1,200 characters by default
- expose a bounded `snippetChars` override (100–4,000 characters)
- clamp legacy limits to integer values from 1–20
- enforce a non-bypassable 50 KiB aggregate output ceiling, including zero-result responses
- report truncation and output-size metadata without duplicating the complete output in `details`

## Root cause

A production Pi session received a 5,787,158-character `session_search` tool result from one indexed message. The result exceeded the model context window even after compaction, while duplicated `details.output` inflated the persisted session. Pi then repeatedly failed overflow recovery and consumed a full CPU core processing the oversized context.

This applies AXI's token-budget principles at the tool boundary: compact defaults, explicit size hints, bounded opt-in detail, and a hard aggregate ceiling on every response path.

## Verification

- regression reproduces a 6,000,096-character response before the fix
- aggregate regression proves 20 large results cannot exceed 50 KiB
- zero-result regression proves an oversized query cannot bypass the ceiling
- focused suite: 7/7 passed
- `npm run check`
- combined hardening branch: all 38 test files passed on Node 24.16 and Node 26

### DIFF
--- a/src/tools/session-search-tool.ts
+++ b/src/tools/session-search-tool.ts
@@ -14,6 +14,10 @@ interface SearchResult {
   count?: number;
   message?: string;
   output?: string;
+  outputChars?: number;
+  outputTruncated?: boolean;
+  snippetChars?: number;
+  truncatedCount?: number;
   ranges?: SessionAnchorRange[];
 }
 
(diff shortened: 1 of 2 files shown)


chandra447:
Merged — thanks for the fix and the thorough regression coverage (oversized-response, zero-result, and aggregate-ceiling cases). Great example of catching a real production failure mode and turning it into a solid, well-tested guard rail. Appreciate the contribution!

PR #101, opened 2026-07-11
fix: protect isolated child prompts and auth adapters

### LINKED ISSUE #94: Child pi subprocesses strip provider auth adapters (--no-extensions) — Claude Pro/Max OAuth users get "out of extra usage" on consolidation/flush/correction saves
## Environment
- pi-hermes-memory `0.7.23`
- pi `0.80.3`
- Auth: Anthropic OAuth (Claude Max subscription) via `pi-claude-oauth-adapter` (installed as a settings.json package)

## Symptom
`/memory-consolidate` (and any subprocess-based flow: session flush, correction save, background-review fallback) fails with an Anthropic quota error even though the main session works fine on the same Claude Max subscription:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"You're out of extra usage. Add more at claude.ai/settings/usage and keep going."}}
```

## Root cause
`src/handlers/pi-child-process.ts` → `appendOwnExtensionArgs()` spawns every child as:

```
pi -p --no-session --no-extensions -e <pi-hermes-memory/src/index.ts> <prompt>
```

`--no-extensions` (added to save prompt tokens/startup CPU) also strips **provider auth adapters** from settings.json. For Claude Pro/Max subscription users, `pi-claude-oauth-adapter` is what normalizes Anthropic OAuth requests and injects the `x-anthropic-billing-header: cc_version=…; cc_entrypoint=…; cch=…;` block that attributes usage to the subscription. Without the adapter, the child still authenticates with the OAuth token from `auth.json`, but the request no longer qualifies as subscription usage — Anthropic bills it to paid **"extra usage"** instead. If the user has no extra-usage balance, every child call fails with the error above.

(pi core itself acknowledges this class of problem — it ships a warning setting: *"Anthropic extra usage — Warn when Anthropic subscription auth may use paid extra usage"*.)

The existing `retryWithoutOverrides` logic doesn't help: it only matches model/provider-resolution failure text, not billing/quota errors — and retrying without overrides re-runs the same adapter-less invocation anyway.

## Reproduction
```bash
# How the package spawns children today — FAILS on a Max-subscription OAuth setup
pi -p --no-session --no-extensions \
  -e ~/.pi/agent/npm/node_modules/pi-hermes-memory/src/index.ts \
  "Reply with exactly: OK"
# → 400 invalid_request_error: "You're out of extra usage..."

# Same invocation with the auth adapter re-added — WORKS (bills to subscription)
pi -p --no-session --no-extensions \
  -e ~/.pi/agent/npm/node_modules/pi-hermes-memory/src/index.ts \
  -e ~/.pi/agent/npm/node_modules/pi-claude-oauth-adapter/extensions/index.ts \
  "Reply with exactly: OK"
# → OK
```

## Affected code paths
All `execChildPrompt` callers: `auto-consolidate.ts`, `session-flush.ts`, `correction-detector.ts`, and the `background-review.ts` subprocess fallback. (The default `reviewTransport: "direct"` path is separate, but everything that goes through `pi -p` is affected.)

## Suggested fixes
Any of these would work:
1. **`childExtensionPaths` config option** — extra `-e` paths appended to every child invocation, so users can re-add auth adapters (most flexible, no hardcoding).
2. **Auto-detect known auth adapters** — if `pi-claude-oauth-adapter` (or similar) is present in the installed packages, append its extension entry to child args automatically.
3. **Drop `--no-extensions` when the default provider uses subscription OAuth** — heavier startup, but correct by construction.

## Local workaround (validated)
Patched `appendOwnExtensionArgs` in the installed copy to append the adapter when present:

```ts
const AUTH_ADAPTER_EXTENSION_PATHS: string[] = (() => {
  try {
    const nodeModulesRoot = resolve(dirname(fileURLToPath(import.meta.url)), "../../..");
    const adapterPath = join(nodeModulesRoot, "pi-claude-oauth-adapter/extensions/index.ts");
    return existsSync(adapterPath) ? [adapterPath] : [];
  } catch {
    return [];
  }
})();

function appendOwnExtensionArgs(args: string[]): void {
  if (OWN_EXTENSION_PATH) {
    args.push("--no-extensions", "-e", OWN_EXTENSION_PATH);
    for (const adapterPath of AUTH_ADAPTER_EXTENSION_PATHS) {
      args.push("-e", adapterPath);
    }
  }
}
```

After this, consolidation subprocess calls succeed and bill to the subscription.

### LINKED ISSUE #95: Windows Defender false positive/privacy risk: background review fallback passes full memory prompt via argv
## Summary

On Windows 11, Microsoft Defender flagged a `pi-hermes-memory` background review fallback subprocess as `Trojan:Win32/ClickFix.FQ!MTB` / severe. The alert appears to be a heuristic false positive, but it exposed a real Hermes issue: the fallback child `pi` invocation passes the full review payload (current memory, user profile, project memory, and conversation transcript) as a command-line argument.

That creates two problems:

1. **AV/EDR false-positive risk**: a very long `node.exe ... pi ... -e pi-hermes-memory/src/index.ts <natural-language agent prompt>` command line looks suspicious to Defender.
2. **Local privacy/logging risk**: Windows Defender Protection History and other process/EDR logs capture argv, so sensitive memory/profile/project content can be stored in OS/security logs.

## Environment observed

- OS: Windows 11
- `pi-hermes-memory`: `0.7.23`
- `@earendil-works/pi-coding-agent`: `0.80.3`
- Hermes config had `reviewTransport: "direct"`, `reviewEnabled: true`, `llmModelOverride: "openai/gpt-5.4-nano"`, `llmThinkingOverride: "off"`.
- Defender detection time: 2026-07-02 02:18 local
- Defender affected resource was a command line only; no separate payload file was listed in the alert I observed.

## Redacted command shape from Defender

```text
CmdLine: C:\Program Files\nodejs\node.exe
  C:\Users\...\AppData\Roaming\npm\node_modules\@earendil-works\pi-coding-agent\dist\cli.js
  -p --no-session --model openai/gpt-5.4-nano --thinking off
  --no-extensions -e C:\Users\...\.pi\agent\npm\node_modules\pi-hermes-memory\src\index.ts
  "Review the conversation above and consider these aspects: ...
   --- Current Memory --- ...
   --- Current User Profile --- ...
   --- Current Project Memory --- ...
   --- Conversation to Review --- ..."
```

The omitted sections included actual user profile/project memory and recent transcript content.

## Code path

From the installed package:

- `src/handlers/background-review.ts`
  - direct transport is documented as default, but if direct review does not succeed it falls back to subprocess review.
  - `buildSubprocessReviewPrompt(...)` includes current memory, user profile, project memory, and conversation parts in one prompt string.
  - `runSubprocessReview(...)` calls `execChildPrompt(pi, prompt, ...)`.

- `src/handlers/pi-child-process.ts`
  - `buildChildPiPromptArgs(prompt, config, ...)` builds `['-p', '--no-session', ...]`, appends `--no-extensions -e <own extension path>`, then does `args.push(prompt)`.
  - On Windows, `resolveChildPiInvocation(...)` launches via `process.execPath` with `[piCliPath, ...args]`, so the full prompt appears in the `node.exe` command line.

## Expected behavior

Large/sensitive review payloads should not be passed via argv. Prefer one of:

- stdin to child `pi`, if supported;
- a temp file with strict permissions and prompt-path argument, then delete after use;
- an IPC/channel API;
- or keep subprocess fallback disabled unless an argv-safe transport exists.

Ideally Hermes would also log/notify the direct-review fallback reason at debug level, because the config was `reviewTransport: "direct"` but the Defender event shows the subprocess path was reached.

## Actual behavior

When direct review fell back to subprocess review, the full review prompt was appended to argv. Defender captured it in Protection History and classified the command line as `Trojan:Win32/ClickFix.FQ!MTB`.

## Impact

Even if Defender is overreacting, the argv transport is enough to leak sensitive memory/profile/project context into local process/security logs. This is especially risky for a memory extension, because the payload is intentionally a condensed set of durable user/project preferences and recent work context.

## Suggested fix

Change child review/consolidation invocations so prompt payload is transferred out-of-band from argv, and add tests asserting sensitive prompt text is not present in the spawned command arguments.

robinbraemer:
## Summary

- pass private review/consolidation/flush/correction prompts through mode-0600 temporary files instead of argv
- retain `--no-extensions` isolation for child Pi processes
- load only Hermes, explicitly trusted `childExtensionPaths`, and the narrowly detected `pi-claude-oauth-adapter`
- preserve configured auth adapters on both the primary and model-override retry paths
- clean temporary prompts best-effort without turning a completed child result into failure
- resolve the Pi CLI safely on Windows

## Root causes

The fallback subprocess placed complete memory and conversation payloads in the process command line, exposing private content to process inspection and AV/EDR heuristics. Separately, `--no-extensions` removed provider OAuth adapters, so child Pi calls lost the authenticated provider path used by the parent session.

The fix keeps the isolation boundary but makes its trusted dependencies explicit. Arbitrary inherited `-e` extensions are deliberately excluded so they cannot observe private review prompts or reintroduce unrelated tools and side effects.

Closes #94.
Closes #95.

Thanks to @Akshay-patel7 for the auth-adapter diagnosis and @flumignano for the Windows privacy/Defender report.

## Verification

- all 36 upstream test files passed on Node 24.16
- focused child/caller suite: 154 tests passed
- `npm run check`
- regressions cover argv privacy, mode 0600, cleanup on success/failure, cleanup failure, Windows invocation, explicit adapters on both attempts, automatic Claude OAuth discovery, and exclusion of unrelated inherited extensions

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,7 +471,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `sessionSearch` | `{ "variant": "legacy" }` | Session search implementation: `legacy` keeps the existing SQLite/FTS snippet search; `anchors` uses the opt-in Markdown request surface and returns compact JSONL line-range anchors from `~/.pi/agent/sessions/` |
 | `llmModelOverride` | unset | Optional model override for background review (direct and subprocess), correction save, session flush, and consolidation |
 | `llmThinkingOverride` | unset | Optional thinking override for those LLM calls; valid values are `off`, `minimal`, `low`, `medium`, `high`, and `xhigh`. If `llmModelOverride` is set and this is omitted, review/child calls default to `off` |
-| `childExtensionPaths` | unset | Extra extension entry paths required by isolated child Pi processes, such as provider auth adapters; the installed `pi-claude-oauth-adapter` is detected automatically |
+| `childExtensionPaths` | unset | Trusted provider/auth adapter entry paths explicitly allowed in isolated child Pi processes; the installed `pi-claude-oauth-adapter` is detected automatically |
 | `nudgeInterval` | `10` | Turns between auto-reviews |
 | `nudgeToolCalls` | `15` | Tool calls between auto-reviews (OR with turns) |
 | `reviewRecentMessages` | `0` | Recent messages included in background review (`0` = all) |
--- a/README.md
+++ b/README.md
@@ -471,6 +471,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `sessionSearch` | `{ "variant": "legacy" }` | Session search implementation: `legacy` keeps the existing SQLite/FTS snippet search; `anchors` uses the opt-in Markdown request surface and returns compact JSONL line-range anchors from `~/.pi/agent/sessions/` |
 | `llmModelOverride` | unset | Optional model override for background review (direct and subprocess), correction save, session flush, and consolidation |
 | `llmThinkingOverride` | unset | Optional thinking override for those LLM calls; valid values are `off`, `minimal`, `low`, `medium`, `high`, and `xhigh`. If `llmModelOverride` is set and this is omitted, review/child calls default to `off` |
+| `childExtensionPaths` | unset | Extra extension entry paths required by isolated child Pi processes, such as provider auth adapters; the installed `pi-claude-oauth-adapter` is detected automatically |
 | `nudgeInterval` | `10` | Turns between auto-reviews |
 | `nudgeToolCalls` | `15` | Tool calls between auto-reviews (OR with turns) |
 | `reviewRecentMessages` | `0` | Recent messages included in background review (`0` = all) |
--- a/src/config.ts
+++ b/src/config.ts
@@ -141,6 +141,12 @@ export function loadConfig(configPath = DEFAULT_CONFIG_PATH): MemoryConfig {
         if (trimmed.length > 0) config.llmModelOverride = trimmed;
       }
       if (isThinkingLevel(parsed.llmThinkingOverride)) config.llmThinkingOverride = parsed.llmThinkingOverride;
+      if (isStringArray(parsed.childExtensionPaths)) {
+        const childExtensionPaths = [...new Set<string>(
+          (parsed.childExtensionPaths as string[]).map((item) => item.trim()).filter(Boolean),
+        )];
+        if (childExtensionPaths.length > 0) config.childExtensionPaths = childExtensionPaths;
+      }
       if (hasMemoryOverflowStrategy) {
         config.autoConsolidate = config.memoryOverflowStrategy === "auto-consolidate";
       } else if (hasLegacyAutoConsolidate) {
--- a/src/handlers/pi-child-process.ts
+++ b/src/handlers/pi-child-process.ts
@@ -5,8 +5,9 @@ import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import type { MemoryConfig, ThinkingLevel } from "../types.js";
+import { AGENT_ROOT } from "../paths.js";
 
-type ChildLlmConfig = Pick<MemoryConfig, "llmModelOverride" | "llmThinkingOverride">;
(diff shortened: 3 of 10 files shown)


chandra447:
Merged — thanks for tackling both the auth-adapter isolation and the argv privacy leak in one well-scoped fix. Solid root-cause work on the Windows shim resolution too. Really appreciate the fix and the collaboration getting CI green.

Issue #96, opened 2026-07-06
Corrupt sessions.db triggers unbounded rebuild fork-storm (no lock/backoff/circuit-breaker)


paxcalpt:
## Summary

When the memory database (`~/.pi/agent/pi-hermes-memory/sessions.db`, SQLite in WAL mode) becomes corrupt, the corruption-recovery path escalates a single bad file into a runaway process storm. On my machine (macOS, Apple Silicon, node 26.3.1) this pushed the 1-minute load average to **87** and produced roughly **196 concurrent Pi processes**, ~776 `sessions.db.corrupt-*` quarantine snapshots, and ~113 `sessions.db.rebuild-*.tmp` files totalling ~1.4 GB, before I killed everything manually.

Version: **pi-hermes-memory 0.7.23**.

## Root cause

The recovery logic appears to have **each Pi process independently** detect the malformed DB, quarantine it (`sessions.db.corrupt-<ts>-<pid>`), and rebuild from scratch (`sessions.db.rebuild-<pid>-<ts>.tmp`), with:

- no cross-process advisory lock guarding the rebuild (every process starts its own),
- no exponential backoff between attempts,
- no circuit breaker after repeated failures, and
- no bound on the number of retained `.corrupt-*` / `.rebuild-*.tmp` files.

With several active Pi sessions each firing background memory writes / auto-consolidation, rebuild workers were spawned faster than they could finish. Rising load slowed each worker, which triggered more retries: a positive feedback loop that only stopped when I killed the process cohort. Notably the rebuilt DB was actually valid at the end (`PRAGMA integrity_check = ok`); the loop was never gated on success.

## How the DB corrupted (trigger, for context)

The initial corruption was environmental, not your bug: the host process holding the WAL-mode DB open was suspended by macOS across repeated battery maintenance-sleep cycles and a power-source-change wake (suspend mid-WAL-checkpoint is a known SQLite corruption vector). The point of this issue is not the one-time corruption; it is that recovery from *any* corruption fork-storms the machine.

## Steps to reproduce (conceptually)

1. Corrupt `~/.pi/agent/pi-hermes-memory/sessions.db` (e.g. truncate the WAL mid-write, or `echo garbage >> sessions.db`).
2. Have one or more active Pi sessions performing memory writes.
3. Observe: each process quarantines + rebuilds, process count and `.rebuild-*.tmp` files climb without bound.

## Suggested fixes (priority order)

[...]

PR #102, opened 2026-07-11
fix: serialize memory mutations and consolidation

robinbraemer:
## Summary

- serialize mutations by canonical Markdown target with a token-owned SQLite lock coordinator
- preserve external edits with compare/replay at the real publish boundary instead of an `mtime:size` check alone
- prevent duplicate same-target consolidation children while allowing distinct stores/targets to proceed
- reload authoritative disk state after every failed add/replace/remove
- preserve displaced and conflicting bytes in bounded, symlink-safe recovery artifacts
- keep published commits successful when cleanup/release fails, with resilient same-process lock release
- canonicalize file aliases, including dangling relative symlinks, so alias/direct writers share one lock and inode

## Provenance

This PR intentionally retains the original commits and Git authorship from:

- #97 by Ricardo Henriques (@paxcalpt)
- #99 by Trofimov Alex (@overpod)

The follow-up commit hardens those fixes against the confirmed races found during adversarial review, including the two Copilot findings on #97 (disappearing-lock retry and assertion-safe test cleanup).

## Root causes

The old store mutated cached arrays and replaced Markdown without owning the target across refresh, publish, and SQLite observation. A separate consolidation process, Pi session, editor, symlink alias, or failed save could therefore be overwritten—or a failed mutation could leak into a later successful one. Consolidation locking also lacked stable storage identity and safe ownership/release semantics.

## Stack

This is stacked on #101 for the already-separated child-process privacy/auth changes. Once #101 merges, GitHub will reduce this PR to the mutation/consolidation commits shown after it.

## Verification

- all 38 test files passed on Node 24.16
- focused mutation/consolidation suite: 93 tests passed
- `npm run check`
- original contributor commits remain intact in history
- regressions cover same-size rewrites, pre/post-publish races, open descriptors, rollback successors, failed mutation leakage, alias convergence, lock token ownership/PID reuse, release failures, cross-store lock identity, and artifact age/count/byte caps

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -7,14 +7,84 @@
  * from disk after consolidation completes.
  */
 
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import { MemoryStore } from "../store/memory-store.js";
 import { CONSOLIDATION_PROMPT, ENTRY_DELIMITER } from "../constants.js";
 import type { ConsolidationResult, MemoryConfig } from "../types.js";
+import { AGENT_ROOT } from "../paths.js";
 import { execChildPrompt } from "./pi-child-process.js";
 
 type MemoryTarget = "memory" | "user" | "failure";
 type ToolMemoryTarget = MemoryTarget | "project";
+
+const CONSOLIDATION_LOCK_STALE_GRACE_MS = 30000;
+const CONSOLIDATION_LOCK_ENV = "PI_HERMES_CONSOLIDATION_LOCK_DIR";
+
+interface ConsolidationLock {
+  path: string;
+  release: () => Promise<void>;
+}
+
+function consolidationLockRoot(): string {
+  return process.env[CONSOLIDATION_LOCK_ENV]?.trim()
+    || path.join(AGENT_ROOT, "pi-hermes-memory", ".consolidation-locks");
+}
+
+function sanitizeLockPart(value: string): string {
+  return value.replace(/[^a-z0-9._-]+/gi, "_").slice(0, 80) || "unknown";
+}
+
+function consolidationLockPath(target: MemoryTarget, toolTarget: ToolMemoryTarget): string {
+  return path.join(consolidationLockRoot(), `${sanitizeLockPart(toolTarget)}-${sanitizeLockPart(target)}.lock`);
+}
+
+async function lockIsStale(lockDir: string, timeoutMs: number): Promise<boolean> {
+  try {
+    const stat = await fs.stat(lockDir);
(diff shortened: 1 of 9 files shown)


chandra447:
Reviewed this + the stacked #103 tip in an isolated worktree before merging (`npx tsc --noEmit` clean, 38/38 test files pass). One confirmed bug worth fixing before merge, one alarm I checked and ruled out:

**Confirmed: `staleMs` is accepted but never enforced in `AtomicLockCoordinator.tryAcquire()`**

`AtomicLockOptions.staleMs` (`src/store/atomic-lock-coordinator.ts:22`, required field) is threaded through from all three call sites:
- `src/store/db.ts:368` — `{ staleMs: this.recoveryOptions.recoveryLockStaleMs }`
- `src/store/markdown-mutation-lock.ts:18,25` — `{ staleMs: MUTATION_STALE_MS }`
- `src/extension-root-migration.ts:130` — `{ staleMs: 300_000 }`

But `tryAcquire()` (`atomic-lock-coordinator.ts:104-141`) never reads `options.staleMs`. The owner row's `acquired_at` is selected (line 115/118) but never compared against `now`/any threshold — takeover is decided purely by `probeIncarnation`/`processIsAlive` (lines 127-141).

Effect: a lock held by a process that's *alive but wedged* (blocked syscall, infinite loop, SIGSTOP, attached debugger — not crashed, not exited) is held **forever**. Waiters just time out after `MUTATION_WAIT_MS`/`recoveryLockWaitMs` and throw; the lock itself is never reclaimed until the wedged process actually dies. That's the same failure class #96 was written to fix, one layer up in the primitive meant to generalize the fix.

`db.test.ts`'s "takes over a stale recovery lock" (~347-365) only seeds a dead PID (999999), so it exercises the liveness path, not time-based staleness — the gap isn't caught by the current suite.

I'd be cautious about a naive fix here: blindly force-stealing from a still-*alive* owner risks split-brain (two processes writing the same Markdown file concurrently) unless writes are also fenced by lock token, which doesn't currently exist on the write side. Worth a deliberate design call rather than a quick patch — happy to help think through it if useful.

**Checked and ruled out**: fifo-eviction writing outside the mutation lock. Traced the call chain — `_add()` (which calls `fifoEvictAndAdd`) is only ever invoked from `addWithConsolidation()`'s `runTargetMutation()` callback (`memory-store.ts:590-595`), so it *is* covered by `withMarkdownMutationLock`. Not an issue.

Also, once #101 lands (merged as #101 → 10fa4d4), this branch + #103 will need a rebase onto the new `main` — GitHub is currently showing this as conflicting against the squashed commit even though the file content should be identical.

chandra447:
Merged — thanks for the memory-mutation/consolidation race fix, and for being receptive to the staleMs lease-takeover hardening we worked through together before merging. Good example of iterating on a review finding instead of just shipping the first pass. Appreciate the contribution!

PR #103, opened 2026-07-11
fix: make SQLite recovery and reconciliation authoritative

### LINKED ISSUE #96: Corrupt sessions.db triggers unbounded rebuild fork-storm (no lock/backoff/circuit-breaker)

### LINKED ISSUE #98: memory_search surfaces orphaned SQLite rows that never self-heal and can't be deleted (Markdown↔SQLite reverse-delete gap)
## Bug

The SQLite search mirror (`sessions.db` `memories` table) can hold rows that have **no corresponding entry in the Markdown source of truth** (`USER.md` / `MEMORY.md` / `failures.md`). Once a row is orphaned this way, it keeps appearing in `memory_search` results, never self-heals, and **cannot be removed by any exposed path**:

- `memory` `remove`/`replace` match `old_text` against the **Markdown** entries only (`MemoryStore.remove`/`replace` → `entriesFor(target).filter(e => stripMetadata(e).includes(oldText))`), so they return `No entry matched '...'`.
- `/memory-sync-markdown` is **import-only**: `syncMemoryEntry` does `INSERT` (new) or `UPDATE` (matched existing) and never deletes SQLite rows absent from Markdown.
- There is no aging/pruning/expiry of SQLite rows, and `deleteMemoryById` (the only unconditional `DELETE FROM memories`) is not exposed to any tool or command.

Net effect: the orphan is permanent from the user/agent's perspective and can only be cleared with manual `sqlite3` surgery.

This is distinct from #52 (fixed in #59 / 0.7.13), which addressed rendered-vs-raw text mismatch for entries **that still exist in Markdown**. Here the entry exists **only** in SQLite, so #52's both-stores remove path never engages.

## How a row gets orphaned

Any operation that changes Markdown without a matching SQLite delete, e.g.:
- Auto-consolidation rewriting/merging entries (`auto-consolidate.ts`) such that the pre-consolidation text no longer exists verbatim in Markdown, while the old FTS row lingers.
- A direct/out-of-band edit or manual cleanup of a `.md` file.
- Any prior sync bug that inserted a row later removed from Markdown.

## Observed (pi-hermes-memory@0.7.23)

`memory_search` returns a stale entry:
```
👤 [global] <stale entry text> ...
```
But it is in no Markdown file:
```
$ grep -rn "<distinctive substring>" <memoryDir>/*.md
(no match)
```
It exists only in SQLite:
```
$ sqlite3 sessions.db "SELECT id,target,project,substr(content,1,60) FROM memories WHERE content LIKE '%<distinctive substring>%';"
2|user||<stale entry text truncated>
```
`memory remove`/`replace` with `target:"user"` and substrings of that text both return `No entry matched '...'`.

Running the documented remedy does **not** fix it — verified on 0.7.23, `/memory-sync-markdown` left the orphaned row in place (it only inserts/updates, never prunes).

## Repro shape

1. Add a memory (syncs to Markdown + SQLite).
2. Remove/alter it directly in the `.md` file (or trigger a consolidation that rewrites it).
3. `memory_search` still returns the old text.
4. `memory remove` with that text → `No entry matched`.
5. `/memory-sync-markdown` → orphan still present.

## Expected behavior

The SQLite mirror should stay consistent with the Markdown source of truth **automatically**, without the user running any command. Search should never surface an entry that no longer exists in Markdown.

## Possible fixes (in rough priority)

1. **Automatic reconciliation** — make writes self-healing. Since Markdown is the source of truth, any mutation (add/replace/remove, consolidation, session flush) should reconcile SQLite so orphans can't accumulate. Options:
   - On consolidation/rewrite, delete the pre-consolidation SQLite rows for that target instead of relying on text-match sync.
   - Periodically (e.g. on `session_start` or after consolidation) diff Markdown vs. SQLite for each target and prune rows with no Markdown counterpart.
2. **Fix the sync primitive** — `syncMarkdownMemoriesToSqlite`/`syncMemoryEntry` currently only insert/update. Add a delete pass so a Markdown→SQLite sync is a true reconcile (removes rows absent from Markdown), and run it automatically rather than only via `/memory-sync-markdown`.
3. **Remove-by-id fallback** — expose the SQLite `id` in `memory_search` output and allow `memory remove`/`replace` by id (issue #52 also suggested stable IDs). This is a manual escape hatch, not the primary fix — automatic reconciliation should mean users never need it.

## Note

`/memory-sync-markdown` exists but (a) is import/update-only so it doesn't prune orphans, and (b) requires the user to know it exists and run it. Self-healing should not depend on a manual command.

## Environment

- Package: `pi-hermes-memory@0.7.23`
- Repo: `chandra447/pi-hermes-memory`

robinbraemer:
## Summary

- reconcile each Markdown target/project scope transactionally so deleted, duplicate, stale-category, and orphaned SQLite rows self-heal
- run reconciliation inside the same canonical Markdown mutation lease as the write
- make SQLite corruption recovery single-owner, token-safe, bounded, and circuit-broken across processes and symlink aliases
- cap quarantine/rebuild artifacts and clean abandoned rebuilds
- migrate legacy SQLite generations through a write-excluded, integrity-checked snapshot/readiness protocol
- preserve corrupt generations for the recovery path and preserve all rollback artifacts when restoration is incomplete
- protect migration publication/rollback from concurrent source or destination successors
- make missing canonical project memory authoritative empty instead of resurrecting retained legacy backups

## Root causes

Markdown is the source of truth, but the old write path only added mirror rows; it never reconciled deletions or stale scope/category identities. Corruption recovery was independently attempted by every Pi process, creating rebuild children and unbounded artifacts. Legacy DB migration also exposed partial generations and could lose WAL commits or rollback over concurrent paths.

Closes #96.
Closes #98.

Thanks to @paxcalpt for the corruption fork-storm reproduction and @niksite for the orphaned mirror-row diagnosis.

## Stack

This is stacked on #102 (which is stacked on #101). The commits are separated by subsystem; after the earlier PRs merge, GitHub will reduce this PR to the SQLite recovery/reconciliation commit.

## Verification

- all 38 test files passed on Node 24.16
- focused SQLite/migration/reconciliation suite: 185 tests passed
- `npm run check`
- regressions cover recovery owner races, PID reuse and stale tokens, symlink/dangling aliases, circuit semantics, artifact caps, Markdown deletion and exact-scope pruning, correction/failure scope metadata, WAL snapshot consistency, concurrent writers/openers, corrupt migration handoff, main-last readiness, rollback ownership, and preserved recovery directories

### DIFF
--- a/src/extension-root-migration.ts
+++ b/src/extension-root-migration.ts
@@ -1,12 +1,41 @@
 import fs from "node:fs/promises";
 import { existsSync } from "node:fs";
+import { randomUUID } from "node:crypto";
 import path from "node:path";
+import Database from "better-sqlite3";
+import { AtomicLockCoordinator, type AtomicLockLease } from "./store/atomic-lock-coordinator.js";
+import { canonicalStoragePathSync } from "./store/canonical-storage-path.js";
+
+const DATABASE_FILES = ["sessions.db", "sessions.db-wal", "sessions.db-shm"] as const;
+const DATABASE_MIGRATION_PENDING_FILE = ".sessions-db-migration-pending";
 
 export interface ExtensionRootMigrationResult {
   moved: number;
   merged: number;
   skipped: number;
   warnings: string[];
+  criticalFailures: Array<{
+    name: string;
+    source: string;
+    target: string;
+    message: string;
+  }>;
+}
+
+export interface ExtensionRootMigrationOptions {
+  moveFile?: (source: string, target: string) => Promise<void>;
+  publishDatabaseFile?: (source: string, target: string) => Promise<void>;
+  retireDatabaseFile?: (source: string, target: string) => Promise<void>;
+  backupDatabase?: (source: string, staged: string, onProgress?: () => void) => Promise<void>;
+  onDatabaseBackupProgress?: () => void;
+}
+
+const MIGRATION_LOCK_WAIT_MS = 5000;
+const MIGRATION_LOCK_POLL_MS = 50;
+
+export function isDatabaseMigrationPending(legacyRoot: string, targetRoot: string): boolean {
+  return existsSync(path.join(targetRoot, DATABASE_MIGRATION_PENDING_FILE))
+    || (existsSync(path.join(legacyRoot, "sessions.db")) && !existsSync(path.join(targetRoot, "sessions.db")));
 }
 
 async function pathExists(filePath: string): Promise<boolean> {
@@ -16,6 +45,24 @@ async function pathExists(filePath: string): Promise<boolean> {
   } catch {
     return false;
   }
+}
+
+async function pathEntryExists(filePath: string): Promise<boolean> {
+  try {
+    await fs.lstat(filePath);
+    return true;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === "ENOENT") return false;
+    throw error;
+  }
+}
+
+async function databaseFilesAt(root: string): Promise<string[]> {
+  const names: string[] = [];
+  for (const name of DATABASE_FILES) {
+    if (await pathEntryExists(path.join(root, name))) names.push(name);
+  }
+  return names;
 }
 
 async function moveFileSafe(source: string, target: string): Promise<void> {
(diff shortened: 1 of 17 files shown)


chandra447:
Merged — thanks for making SQLite recovery and reconciliation authoritative here, and for carrying the staleMs fencing through cleanly after the rebase. Nice, thorough piece of hardening. Appreciate the work!

PR #106, opened 2026-07-12
fix: hard-stop runaway child process trees

### LINKED ISSUE #105: Auto-consolidation can enter an unbounded invalid memory-tool loop
## Summary

Auto-consolidation can launch a non-interactive Pi child that enters an unbounded invalid-tool-call loop. Because the child uses `-p --no-session`, it is easy to miss in normal agent/session monitoring while it continues spending API budget.

## Observed impact

One incident produced:

- **2,118 recorded model turns** over **159 minutes**
- Requests roughly every **3–10 seconds**
- Approximately **1 billion generated tokens** reported by the provider console

No secrets, prompts, API-key identifiers, session IDs, or project data are included here.

## Trigger

With the default configuration in **v0.7.23**:

- `memoryOverflowStrategy: "auto-consolidate"`
- `autoConsolidate: true`
- `consolidationTimeoutMs` configured by the extension defaults

a memory write reached capacity and initiated consolidation. The extension starts a child Pi process using the memory tool to merge entries.

## Failure mode

The child repeatedly attempted a memory replacement/consolidation call without the required `content` parameter. The model output would acknowledge the missing parameter and claim that the next call would include it, but the next call remained invalid. This repeated continuously instead of terminating after a bounded number of invalid tool calls or after the configured timeout.

The extension source appears to pass a 60-second timeout to the child invocation, but the observed child ran for 159 minutes.

## Expected behavior

Auto-consolidation should be fail-closed and bounded:

1. Enforce a hard wall-clock timeout that actually terminates the child and its descendants.
2. Enforce a maximum model-turn / tool-call / token budget for consolidation.
3. Abort after repeated identical validation failures (especially required-argument errors).
4. Surface an explicit failure to the parent agent rather than silently continuing.
5. Make auto-spawned consolidation work observable, even when `--no-session` is used.

## Suggested regression test

Mock a provider that repeatedly emits an invalid `memory.replace` call missing `content`. Assert that consolidation exits after a small bounded number of attempts and that no further provider requests are issued.

robinbraemer:
## Summary

- run every non-interactive child Pi behind a bundled process-tree watchdog
- enforce the configured wall-clock timeout independently and escalate from graceful termination to a hard tree kill
- use POSIX process groups and Windows `taskkill /T /F` so descendants cannot survive
- route cancellation through a private marker so Windows cannot kill the watchdog before cleanup
- resolve Windows `pi.cmd` shims to the adjacent `cli.js` without shell interpolation
- preserve private prompt files, extension isolation, trusted auth adapters, and override retry behavior
- report timeout/cancellation explicitly to the parent

## Root cause

Pi's extension `execCommand` marks a process as `killed` as soon as `SIGTERM` is sent, then only sends `SIGKILL` when `proc.killed` is false. In Node, that property means the signal was sent—not that the process exited—so a child that ignores or traps `SIGTERM` can continue indefinitely. The old path also targeted only the direct process, not its descendants.

Closes #105.

## Verification

- watchdog regression: a child that ignores `SIGTERM` exits with code 124 within the hard bound
- process-tree regression: a stubborn descendant is also terminated
- cancellation-channel regression: abort terminates the tree without forwarding AbortSignal to the watchdog
- Windows shim and retry regressions
- focused child/consolidation/review/correction/flush suite: 158 tests passed
- full 38-file suite passed on Node 24.16 and Node 26
- `npm run check` and `git diff --check`

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -87,7 +87,7 @@ function describeConsolidationFailure(
   timeoutMs: number,
 ): string {
   const stderr = result.stderr?.trim();
-  const terminated = result.killed || result.code === 143;
+  const terminated = result.killed || result.code === 124 || result.code === 143;
 
   if (terminated) {
     return `Consolidation subprocess was terminated (likely timeout or cancellation). Timeout: ${timeoutMs}ms. Consider increasing consolidationTimeoutMs if this is a manual run.`;
--- a/src/handlers/child-process-watchdog.mjs
+++ b/src/handlers/child-process-watchdog.mjs
@@ -1,0 +1,76 @@
+import { spawn } from "node:child_process";
+
+const [timeoutValue, command, ...args] = process.argv.slice(2);
+const timeoutMs = Number(timeoutValue);
+
+if (!command || !Number.isFinite(timeoutMs) || timeoutMs <= 0) {
+  process.stderr.write("pi-hermes-memory watchdog: invalid invocation\n");
+  process.exit(2);
+}
+
+const child = spawn(command, args, {
+  detached: process.platform !== "win32",
+  stdio: ["ignore", "pipe", "pipe"],
+});
+
+child.stdout?.pipe(process.stdout);
+child.stderr?.pipe(process.stderr);
+
+let timedOut = false;
+let terminating = false;
+let forceTimer;
+
+function signalTree(signal) {
+  if (!child.pid) return;
+  if (process.platform === "win32") {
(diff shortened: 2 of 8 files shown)


chandra447:
Merged — thanks for the process-tree watchdog fix. Root-causing the SIGTERM/`proc.killed` semantics gap and backing it with real spawn/kill regression tests (including the Windows taskkill path) made this an easy merge. Appreciate the fix and the reciprocal back-and-forth throughout this batch of PRs!

PR #110, opened 2026-07-17
fix: wait for concurrent SQLite writers

ShellWen:
## Summary

- configure a 5-second SQLite `busy_timeout` before WAL and schema setup
- add a regression test with a second process holding the write lock

## Why

Every Pi process opens the same `sessions.db`. Live session indexing, startup backfill, and memory synchronization can therefore overlap across processes. WAL permits concurrent readers but still has a single writer, and Bun SQLite defaults `busy_timeout` to `0`, so the losing writer immediately reports:

```
⚠️ Live session indexing failed: database is locked
```

The timeout is connection-local and applies to both the Bun compatibility path and `better-sqlite3`, allowing short write transactions to serialize instead of dropping the indexing attempt.

## Validation

- `npm run check` on Node 24
- `npm test` — all 38 test files passed
- manual Bun 1.3.13 contention check — a write waited for a separate SQLite writer and completed successfully

### DIFF
--- a/src/store/db.ts
+++ b/src/store/db.ts
@@ -69,6 +69,7 @@ class DatabaseCorruptionError extends Error {
   }
 }
 
+export const SQLITE_BUSY_TIMEOUT_MS = 5000;
 export const SQLITE_WAL_AUTOCHECKPOINT_PAGES = 1000;
 
 const DATABASE_FILE_SUFFIXES: readonly DatabaseFileSuffix[] = ['', '-wal', '-shm'];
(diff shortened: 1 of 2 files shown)


chandra447:
Superseded by #113, which landed the same `PRAGMA busy_timeout = 5000` change on main (with a hardened concurrent-writer regression test) along with the #112 disk-authoritative memory mutation fix.

Thanks @ShellWen — the approach here was correct; we just couldn't pull the fork branch cleanly and folded the equivalent change into #113.

chandra447:
Thank you again for this PR — the `busy_timeout` approach was exactly right.

We’re grateful for the contribution. The equivalent change shipped in **v0.8.2** via #113 (with a hardened concurrent-writer regression test), since the fork branch wasn’t available to merge directly. Appreciate you improving multi-process SQLite reliability for everyone using the extension.

PR #118, opened 2026-07-22
fix: rank memory_search by BM25 and tighten NL OR fallback

cat0825:
## Problem

`memory_search` used the FTS5 OR-fallback ordered by `last_referenced` only, so natural-language agent queries like `memory search 检索 无关` returned long, recently-touched notes that merely contained "memory" above actual matches.

## Fix

- **BM25 ranking**: FTS hits are ranked by `bm25(memory_fts)` first, then `last_referenced`, via a JOIN. Relevance beats recency: a memory that is *the* right answer but hasn't been touched in months outranks a fresher weak match, and recency only breaks BM25 ties. This is asserted by the new `old dense match outrank a fresh diluted match` test.
- **Stopword filtering**: EN/ZH stopwords and weak tokens are dropped before MATCH; the OR fallback uses only significant terms and post-filters hits by term coverage (`passesTermHitFilter`).
- **`session_search` LIKE fallback** prefers significant terms so glue words don't expand into unrelated transcript hits; escaped punctuation substrings (e.g. `%`) still work.

## Rebased over v0.8.3 — reconciliation with #127, addressing review

Review points from @chandra447, in order:

**1. Stopword recall loss (blocking)** — resolved by *changing the behavior*, not just pinning it. Hard-empty for all-stopword queries turned out to conflict directly with the just-merged #122 regression test: `find` and `use` are stopwords, so `DO NOT USE FIND /` (and its lowercase spelling) would have returned nothing again. All-stopword queries now degrade to coverage-filtered matching instead of hard-empty: AND over all word terms first, then OR with the term-coverage post-filter. A memory genuinely about "memory search" stays reachable by its own words (pinned by test), while the motivating noise case — a long note casually mentioning "memory" once — stays filtered (also pinned).

**2. CJK path** — `searchMemories` gains a LIKE safety net (coverage-filtered, recency-ordered, only reached when every FTS attempt is empty), because unicode61 tokenizes a CJK run as one token and `名字` can never FTS-MATCH inside `用户的名字是鸣人`. Pinned by a CJK test resolving through that net.

**3. BM25 vs recency** — sentence above + dedicated test.

**4. Interaction with #127's recovery (blocking)** — the parse-error retry and the all-stopword case now share one `runNaturalLanguageRecovery` helper: AND over recovered terms, then the OR fallback with the same BM25 ordering and coverage post-filter as the regular fallback, then the LIKE net. Ranking cannot depend on how the query was spelled. Valid operator queries that match nothing keep exact semantics — recovery only triggers on a parse error.

**quoteTerms folding** — done; the duplicated escaping helpers are gone.

## Testing

- `tests/run-all.sh`: all 41 test files pass; `tsc --noEmit` clean.
- New/updated tests: all-stopword reachability + noise filtering, CJK substring resolution, BM25-before-recency ordering, uppercase-operator recovery (from #127, still green on the merged semantics).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Reload no longer waits for an unnecessary LLM memory flush**: Pi preserves the current session during `/reload`, but the shutdown handler ignored `event.reason` and awaited direct completion plus possible subprocess fallback after six user turns. Reload now skips both transports; quit and session-replacement reasons keep the existing flush policy.
 
+- **`memory_search` ranks by relevance instead of recency** ([#118](https://github.com/chandra447/pi-hermes-memory/pull/118)): FTS5 matches were ordered by `last_referenced DESC` alone, so a long note that mentions a query term once outranked a short memory that is entirely about it, purely because it had been touched more recently. The match now joins `memory_fts` and orders by `bm25(memory_fts)` ascending with `last_referenced DESC` as the tie-break, so the densest match wins and recency only decides between equally relevant rows. Result *sets* are unchanged; only their order is.
+
 ## [0.9.4] - 2026-08-08
 
 ### Added
--- a/src/store/sqlite-memory-store.ts
+++ b/src/store/sqlite-memory-store.ts
@@ -22,6 +22,13 @@ const MEMORY_SELECT_COLUMNS = `
   last_referenced
 `;
 
+// The BM25-ranked search joins memory_fts, which also has a `content` column,
+// so that one query needs the same list qualified with the `memories` alias.
+const MEMORY_SELECT_COLUMNS_M = MEMORY_SELECT_COLUMNS
+  .split(',')
+  .map((column) => `m.${column.trim()}`)
+  .join(',\n        ');
+
 const FAILURE_CATEGORY_SET = new Set<MemoryCategory>([
   'failure',
   'correction',
@@ -703,7 +710,7 @@ export function searchMemories(
   const conditions: string[] = [];
   const params: unknown[] = [];
 
-  // FTS5 match via subquery with escaped query
+  // FTS5 match via JOIN with BM25 ranking
   const normalizedQuery = normalizeFts5Query(query);
   if (normalizedQuery.length === 0) {
     return [];
@@ -715,7 +722,7 @@ export function searchMemories(
     const conditions: string[] = [];
     const params: unknown[] = [];
 
-    conditions.push('m.id IN (SELECT rowid FROM memory_fts WHERE memory_fts MATCH ?)');
+    conditions.push('memory_fts MATCH ?');
     params.push(matchQuery);
 
     if (project !== undefined) {
@@ -740,10 +747,13 @@ export function searchMemories(
     const whereClause = conditions.length > 0 ? `WHERE ${conditions.join(' AND ')}` : '';
 
     const sql = `
-      SELECT ${MEMORY_SELECT_COLUMNS}
+      SELECT
+        ${MEMORY_SELECT_COLUMNS_M},
+        bm25(memory_fts) AS rank_score
       FROM memories m
+      JOIN memory_fts ON memory_fts.rowid = m.id
       ${whereClause}
-      ORDER BY m.last_referenced DESC
+      ORDER BY rank_score ASC, m.last_referenced DESC
       LIMIT ?
     `;
 
@@ -759,6 +769,7 @@ export function searchMemories(
         corrected_to: string | null;
         created: string;
         last_referenced: string;
+        rank_score: number;
       }>;
 
       return rows.map(mapRow);
--- a/tests/store/sqlite-memory-store.test.ts
+++ b/tests/store/sqlite-memory-store.test.ts
@@ -456,6 +456,45 @@ describe('sqlite-memory-store', () => {
       const results = searchMemories(dbManager, 'AND OR NOT');
       assert.strictEqual(results.length, 0);
     });
+
+    it('ranks a dense old match above a fresher, more diluted one (BM25 before recency)', () => {
+      addMemory(dbManager, 'deploy target is the staging cluster');
+      addMemory(
+        dbManager,
+        'unrelated retrospective notes that mention deploy once among a great many other words about hiring, budgets, roadmaps, vendor calls, office moves and quarterly planning for the staging of a conference'
+      );
+      const db = dbManager.getDb();
+      db.prepare("UPDATE memories SET last_referenced = '2025-01-01' WHERE content LIKE 'deploy target%'").run();
+      db.prepare("UPDATE memories SET last_referenced = '2026-08-30' WHERE content LIKE 'unrelated retrospective%'").run();
+
+      const results = searchMemories(dbManager, 'deploy staging');
+
+      // Both rows match; recency ordering would put the long fresh note first.
+      assert.strictEqual(results.length, 2);
+      assert.ok(results[0].content.startsWith('deploy target is the staging cluster'));
+    });
+
+    it('still broadens a two-term natural-language query through the OR fallback', () => {
+      // Neither "dark" nor "chocolate" co-occur, so the implicit-AND query
+      // misses and only the OR fallback can recover this row. The existing
+      // fallback test uses three terms; this pins the two-term case.
+      addMemory(dbManager, 'user prefers dark mode UI theme', 'user');
+
+      const results = searchMemories(dbManager, 'dark chocolate', { target: 'user' });
+
+      assert.ok(results.some((r) => r.content.includes('dark mode')));
+    });
+
+    it('recovers a natural-language query whose raw FTS5 form fails to parse', () => {
+      // "DO NOT USE FIND /" passes through as raw FTS5 syntax and throws;
+      // runSearch must catch that and let the natural-language retry run.
+      addMemory(dbManager, 'Never search whole filesystem from root. Do not run find /.', 'user');
+
+      assert.doesNotThrow(() => searchMemories(dbManager, 'DO NOT USE FIND /', { target: 'user' }));
+      const results = searchMemories(dbManager, 'DO NOT USE FIND /', { target: 'user' });
+
+      assert.ok(results.some((r) => r.content.includes('find /')));
+    });
   });
 
   describe('getMemories', () => {


cat0825:
Status update: rebased check — the branch is already up to date with current `main` (v0.8.2, 4ae0483), no conflicts. Re-ran the full suite locally today: all 40 test files pass (`tests/run-all.sh`), `tsc --noEmit` clean.

Search-quality context: this complements #127 (which fixes the parse-error path from #122) — #118 fixes the ranking side, where OR-fallback results are ordered by `last_referenced` only, so NL queries return recently-touched but barely-relevant notes above actual matches. Whichever PR lands second I'll rebase promptly.

Happy to adjust anything — splitting the stopword filter from the BM25 ordering into separate commits, for example, if that makes review easier.

chandra447:
@cat0825 — status update, since your #127 changed the ground under this one.

#127 is **merged** and ships in 0.8.3. As you predicted, it touches all three of the files here (`fts-query.ts`, `sqlite-memory-store.ts`, `session-search.ts`), so this needs the rebase you offered. Thanks for calling that out up front — it made the ordering decision easy.

**Why #127 went first,** so the sequencing doesn't read as a judgement on this PR: #122 was a reported correctness bug with a reproducible "correct memory exists, search returns nothing" case, and its fix was 94 lines keyed on a single failure signal. This PR is a *precision/ranking* change — larger, more valuable, and with a much wider behavioural blast radius. Landing the narrow bug fix first keeps the two effects separable if either needs revisiting.

**What I'll be looking at on the rebase**, so there are no surprises:

1. **Stopword filtering and recall loss.** "All-stopword NL normalizes to empty (no MATCH)" is the part I want to convince myself about. `memory search related context` returning 0 is clearly right. But a user memory that genuinely *is* about memory search — plausible in this project — becomes unreachable by its own words. I'd like a test pinning what happens when the stopword list eats a query whose target legitimately contains those terms.
2. **The CJK path.** You're candid that this leans on significant terms plus substring hit filtering under the default tokenizer. Since `检索`/`无关` are in the stopword set, I want to see a CJK case that still resolves — ideally the `name identity Naruto`-style example from your table, but in Chinese.
3. **BM25 versus `last_referenced`.** Ranking by relevance first is right. Worth a sentence in the PR body on what this does to the recency-driven cases — a memory that is *the* right answer but hasn't been touched in months should still win over a fresher weak match, and BM25-then-recency gives that, but I'd like it asserted somewhere.
4. **Interaction with #127's recovery path.** `searchMemories` now has a parse-error retry that reruns `runSearch` with a natural-language query. Please make sure the BM25 JOIN and the OR over-fetch/hit-ratio filter apply on that retry path too, not just the first attempt. That's the one place a mechanical rebase could quietly produce two different ranking behaviours depending on how the query was spelled.

Also worth folding in while you're here: `normalizeFts5Query` and `buildFallbackFts5Query` still inline the quote-escaping that your `quoteTerms` helper from #127 now does.

No rush, and no pressure to take all four — 1 and 4 are the ones I'd call blocking. CI on this branch is green as of the last run; I've approved the workflow so future pushes run automatically without waiting on me.

cat0825:
Rebased over v0.8.3 and pushed (`4177e2b`). All four review points addressed — details in the updated PR body, but the short version:

- **1 & 4 (blocking)**: resolved together. While reconciling I found they intersect: `find` and `use` are in the stopword list, so a mechanical rebase would have silently regressed the just-merged #122 fix — `DO NOT USE FIND /` *and* its lowercase spelling would both have gone back to returning nothing. All-stopword queries now degrade to coverage-filtered matching (AND → filtered OR) instead of hard-empty, and the #127 parse-error retry shares that exact recovery path, so ranking is identical regardless of spelling. Both behaviors are pinned by tests.
- **2**: `searchMemories` gains a coverage-filtered LIKE safety net for CJK (unicode61 tokenizes CJK runs whole, so FTS alone can't resolve `名字` against `用户的名字是鸣人`). Pinned by a CJK test.
- **3**: sentence added to the PR body + a dedicated `old dense match outranks fresh diluted match` test.
- `quoteTerms` folding done.

Full suite: 41 test files pass, `tsc --noEmit` clean. The all-stopword semantics change (empty → coverage-filtered) is the one judgement call I made beyond the letter of the review — happy to split it out or revert to hard-empty + pinning test if you'd rather keep this PR narrower.

chandra447:
@cat0825 — the rebase is good work, and catching that `find` and `use` are stopwords (so a mechanical rebase would have silently un-fixed #122) is exactly the kind of thing I was hoping the review would surface. You found an interaction I'd only guessed at.

**Verdict: this is worth landing.** BM25-before-recency is a genuine defect fix, not a nice-to-have — ordering search results by *when they were last touched* is simply the wrong answer for a search tool, and no amount of good memory content compensates. That half I'd take on its own.

I've pushed two fixes rather than sending you round a third time. Both passed CI and your own tests, which is the part worth paying attention to.

### 1. Precision regression — the coverage filter counts substrings

`countTermHitsInText` used a bare `includes()`. That is strictly *looser* than the FTS match it post-filters: unicode61 tokenizes `uses` as one token, so `MATCH "use"` never hits it — but `"uses".includes("use")` does.

Recovery terms for `DO NOT USE FIND /` are `["DO","USE","FIND"]`. Against `deploy uses docker compose and kubernetes manifests`:

- `"do"` → inside **do**cker ✓
- `"use"` → inside **use**s ✓
- `"find"` ✗

2 of 3, threshold `ceil(3 × 2/3)` = 2. **Passes.** So:

```
searchMemories('DO NOT USE FIND /', { target: 'memory' })
  before this PR → []
  with this PR   → [ 'deploy uses docker compose and kubernetes manifests' ]
```

An unrelated memory admitted by a filter whose entire job is rejecting exactly that. Short terms are where it bites, and NL recovery produces short terms.

ASCII word terms are now matched on word boundaries. CJK runs and punctuation have no boundary to anchor on and keep the substring test. This also makes the filter agree with the FTS layer it sits on top of, which it previously contradicted.

### 2. The LIKE net was unreachable for single-term queries

```ts
const fallbackQuery = buildFallbackFts5Query(query);
if (!fallbackQuery || fallbackQuery === normalizedQuery) {
  return exactResults;        // ← LIKE net never reached
}
...
return runLikeSearch(significantTerms);
```

`buildFallbackFts5Query` returns `null` for a single term, so that branch returns early. Your CJK test passes because it uses a multi-term query — but a lone CJK term is the *most natural* way to search Chinese, and it's precisely what the net exists for:

```
searchMemories('名字', { target: 'user' })   // 用户的名字是鸣人 present
  → []   (before my fix)
  → hit  (after)
```

That branch now falls through to the net. I also restricted the net to terms containing non-ASCII characters — its documented purpose. Otherwise every ordinary empty ASCII search pays for a full table scan and re-admits the loose substring hits point 1 is about.

### Also

`fts-query.ts` came back converted wholesale to tabs; the repo is 2-space throughout. I've reverted the indentation so the diff shows your actual logic. Worth checking your editor config before the next one.

### Verification

Independent adversarial probe, all 16 cases behaving:

| query | filter | result |
|---|---|---|
| `DO NOT USE FIND /` | `user` | finds `find /.` |
| `do not use find /` | `user` | same — spelling-independent, as you intended |
| `DO NOT USE FIND /` | `memory` | **0** (was the false positive) |
| `名字` | `user` | finds `用户的名字是鸣人` (was 0) |
| `名字` | `memory` | 0 — filter respected on the net |
| `memory search related context` | — | 1, the genuinely-about-it note; the long casual-mention note filtered |
| `pnpm OR AEST` / `pnpm AND nonexistent` / `AND OR NOT` | — | unchanged |

Two regression tests added for the above. 41 test files pass, `tsc --noEmit` clean.

### Not merging quite yet

Both fixes change *your* design intent — scoping the net and altering match semantics — and this touches every `memory_search` call the agent makes. You've been fast and accurate all day; you've earned a look at the correction before it lands rather than finding it in the changelog.

Take a look and tell me if either lands wrong, particularly the non-ASCII gate on the net — if you had a non-CJK case in mind for it, say so and I'll widen it back with a word-boundary guard instead. Otherwise I'll merge this in a day or so.

On your open question — no, don't split the all-stopword change out. You're right that it's entangled with #122 and separating them would ship a knowingly-broken intermediate state.

cat0825:
Reviewed fbfecf0 locally. Both corrections match the intended behavior. ASCII word-boundary matching keeps the coverage filter aligned with FTS token semantics, and the non-ASCII gate keeps the LIKE fallback scoped to the tokenizer gap it is meant to cover. I did not have an additional ASCII or non-CJK use case in mind for that net. I also fast-forwarded the branch and reran the full verification: all 41 test files pass and `npm run check` is clean. Good to merge from my side.

chandra447 (changes requested):
Apologies for the long silence on this — three months open is not acceptable and that's on me, not you. Thank you for the work, and for responding to the earlier round of comments.

**The BM25 half is correct and I want it.** I checked it properly rather than taking it on faith: `memory_fts` is a single-column FTS5 table so bare `bm25(memory_fts)` has the right arity and default weight, `ORDER BY rank_score ASC` is the correct direction, `EXPLAIN QUERY PLAN` confirms the JOIN still drives off the FTS index (`SCAN memory_fts VIRTUAL TABLE INDEX 0:M1`), every value is bound, and `snippetChars` is untouched. A dedicated security pass also came back clean: no injection, the MATCH expression is a bound parameter, the new regexes are anchored single-class with no ReDoS, and `runSearch` now *catches* FTS5 parse errors that main let through.

**The stopword/coverage half regresses recall in four measurable ways.** All four were reproduced against real indexed databases, main vs this branch:

### 1. `session_search` returns nothing for short tokens and single-char CJK

`collectLikeTerms` (`session-search.ts:49-55`) now returns `collectSignificantSearchTerms(query)`, and when that's empty it keeps only punctuation tokens — so the query short-circuits to `executeSearch({type:'like', terms: []})`, which returns `[]`.

```
MAIN {"go":1,"ts":1,"s3":1,"db":1,"书":1,"Prisma":1,"go mod":1,"bucket":1}
PR   {"go":0,"ts":0,"s3":0,"db":0,"书":0,"Prisma":1,"go mod":1,"bucket":1}
```

`书` going 1→0 directly regresses `b0ead25` (FTS5→OR→LIKE fallback for CJK, #79) — that LIKE net exists precisely so a CJK term unicode61 can't tokenize stays findable by substring.

### 2. The LIKE net drops the ASCII half of a mixed query and returns unrelated rows

`runLikeSearch` filters to non-ASCII terms (`sqlite-memory-store.ts:805`) but then computes the coverage threshold over that *reduced* set (`:867`). `passesTermHitFilter` derives `needed` from `terms.length`, so a two-term mixed query collapses to a one-term requirement and the ASCII constraint vanishes instead of failing the row:

```
MAIN  '名字 kubernetes' -> []
PR    '名字 kubernetes' -> ['用户的名字是鸣人']
```

That result contains no trace of `kubernetes`.

### 3. Sub-3-char ASCII tokens are discarded — and BM25 then ranks the wrong row first

`isSignificantSearchTerm` (`fts-query.ts:329-331`) rejects any latin/numeric token under 3 chars. In a coding-agent memory store those are `go`, `ts`, `js`, `db`, `ci`, `s3`, `v2`, `k8` — usually the only discriminating word:

```
'go build' -> "build"     's3 bucket' -> "bucket"     'v2 api' -> "api"
```

Combined with the new ranking:

```
MAIN 's3 bucket' -> ['artifacts go to the s3 bucket named releases']
PR   's3 bucket' -> ['the local bucket cache lives in /tmp/bucket',   <-- first
                     'artifacts go to the s3 bucket named releases']
```

A precision PR making precision worse on a plain two-word query is the headline risk. Gating the length rule on alphabetic-only tokens, or exempting tokens containing a digit, would fix it.

### 4. `minTermHitRatio(2) === 1` makes the OR fallback dead for two-term queries

`buildFallbackFts5Query` still emits `"a" OR "b"`, but `passesTermHitFilter` then demands both — which is the AND query that already returned nothing.

```
'dark chocolate'    MAIN ['user prefers dark mode UI theme']  PR []
'docker swarm'      MAIN ['deploy uses docker compose']       PR []
'theme preferences' MAIN ['user prefers dark mode UI theme']  PR []
```

That's the behaviour `793d32c` (#65) exists to provide. The existing regression test uses three terms (`name identity Naruto`) so it doesn't cover the two-term case, and this PR doesn't add one. If the tightening is deliberate it needs a test pinning the new semantics and a CHANGELOG line; otherwise ratio `0.5` at `termCount === 2` restores #65 without hurting three-plus-term coverage.

**None of these four is covered by the added tests**, which is why `npm run check` is clean and 41/41 files pass. The tests pin the cases you chose; they'd all still pass with the regressions in place.

## Also worth fixing

- **The stopword list swallows this project's own vocabulary**: `context` (:146), `question` (:153), `search` (:160), `memory` (:163), plus `find`, `use`, `show`, `need`, `want`, and `问题`/`内容` on the Chinese side. `'context window' -> "window"`, `'memory limit' -> "limit"`. I raised partial-stopwording in the earlier round and the response only addressed the all-stopword case — multi-term queries never reach that recovery path, so this narrowing is unrecoverable. Trim the domain nouns and keep only syntactic glue.
- **Duplicate LIKE escaping** (`sqlite-memory-store.ts:815`) beside the existing `escapeLikePattern` at `:193` in the same file. Byte-identical today; a second copy of an escape routine is how the two drift. Call the helper.
- **Dead branch**: `session-search.ts:236-238` `else if (explicitOperatorQuery) return []` is unreachable — `normalizeFts5Query` returns the trimmed input verbatim when an explicit operator is present, and blank queries already returned at `:111`.
- **Duplicates in the stopword literal**: `"of"` at both `:44` and `:102`, `"一些"` twice.
- **Style churn**: `fts-query.ts` was converted wholesale to double quotes (293 occurrences in a file that had none), which is why that file's diff is `+472/-30` for a much smaller logic change and why every future `git blame` there points here. Separately, the new `tests/store/fts-query.test.ts` is still tab-indented (56 of 67 lines) — I flagged the tab conversion last round and `fts-query.ts` got fixed, but the new test file didn't.
- **CHANGELOG**: reformats already-released 0.8.x sections (blank lines after bold headers, table separator restyled). Please keep just the `## [Unreleased]` entry — it also happens to be the one file that conflicts with main.
- **Latent**: `fetchLimit = Math.min(Math.max(limit * fetchMultiplier, limit), 100)` (`:743`) caps the *caller's* limit at 100, not just the over-fetch. No live caller hits it (the tool clamps to 20) but the cap belongs on the multiplier.
- **Latent**: the LIKE net fetches `ORDER BY last_referenced DESC` and filters *after* (`:855-869`), so an old exact match outside the 40 most-recent rows is dropped — the same recency-over-relevance failure this PR exists to remove, left on the one tier with no BM25 score.

## Conflicts

Only one file: `CHANGELOG.md`. Dropping the cosmetic hunks above will likely make the rebase trivial.

## Where I'd like to land

I'd very much like to merge the BM25 ranking. If you're willing, the cleanest path is to **split this into two**: the BM25 change (plus the FTS5 parse-error catch, which is a genuine improvement) goes in essentially as-is, and the stopword/coverage/significant-term work becomes a second PR where we can argue the recall trade-offs with the four cases above as the test set.

If you'd rather not carry it further after this long a wait, say so and I'll take the BM25 half forward myself with attribution to you. Either is completely fine — you've already done the hard part.

cat0825:
Rebased onto current `main` (9 commits ahead, incl. #147/#150) and pushed (`cd4e12b`). Only conflict was the CHANGELOG — merged the PR entry into `[Unreleased]`'s Fixed section with the 0.9.3/0.9.2 entries intact.

Verified locally before pushing: full suite (43 test files) passes, `tsc --noEmit` clean. CI on the updated head is green across all four jobs (Lint, Type Check, Tests, Min Supported SDK).

PR is mergeable (`CLEAN`) whenever you're ready.

chandra447:
Thank you for the careful BM25 work, adversarial testing, and follow-up corrections. The BM25 ranking itself is valuable and the CI is green, but I am not merging this combined PR: the stopword, coverage-filter, CJK LIKE fallback, and short-token behavior changes have a much wider search-semantics blast radius than the ranking fix. Please split the narrow BM25/parse-error work from the broader semantics changes, or provide an explicit follow-up scope for those behaviors.

chandra447:
Review verdict: do not merge this combined PR. The BM25-before-recency and FTS5 parse-error handling are valuable and should be preserved, but the stopword, coverage, short-token, CJK LIKE fallback, and session-search changes have a much wider search-semantics blast radius. The PR is also currently marked CHANGES_REQUESTED/DIRTY in GitHub despite the older comment claiming a clean rebase. Please split or prepare a narrow rebased BM25 PR first, with its focused regression tests, then handle CJK/tokenizer semantics separately under issue #174. This keeps a good ranking fix reviewable without shipping unreviewed recall changes.

chandra447:
Thanks for the detailed BM25, stop-word, CJK, and fallback work. The current branch is stale and conflicts with `main`, so it is not merge-ready. Rebase onto current `main`, resolve the conflicts, and rerun the full typecheck and test suite on the new head. After that update, I will review the ranking and fallback semantics against the current search implementation.

cat0825:
Thanks for the detailed review, and no need to apologise about the wait — the reproductions you did made this easy to act on.

**I've taken your split.** This branch is now the BM25 ranking and nothing else. It is a single commit on top of `main` at 671a27f (0.9.7), and the whole diff is **3 files, +56/-4**:

```
CHANGELOG.md                            |  2 ++
src/store/sqlite-memory-store.ts        | 19 ++++++++++++----
tests/store/sqlite-memory-store.test.ts | 39 +++++++++++++++++++++++++++++++++
```

The ranking change is byte-for-byte the shape you verified: `memory_fts MATCH ?` in the WHERE, `JOIN memory_fts ON memory_fts.rowid = m.id`, bare `bm25(memory_fts) AS rank_score`, `ORDER BY rank_score ASC, m.last_referenced DESC`, `limit` still bound. `snippetChars` untouched. The only thing I had to add beyond your version is a `MEMORY_SELECT_COLUMNS_M` derived constant — the JOIN brings a second `content` column into scope, so that one query needs the column list qualified with the `m.` alias. It's derived from the existing `MEMORY_SELECT_COLUMNS` rather than duplicated, so the two can't drift.

I re-ran your query plan check on the narrowed branch and it's unchanged:

```
SCAN memory_fts VIRTUAL TABLE INDEX 0:M1
SEARCH m USING INTEGER PRIMARY KEY (rowid=?)
USE TEMP B-TREE FOR ORDER BY
```

## Your four regressions

I didn't want to just assert these were gone, so I built a probe with rows modelled on your cases and diffed the narrowed branch against a clean `upstream/main` worktree, same DB contents both sides. **All 12 of your listed queries return identical result _sets_ on both.** Per case:

**1. `session_search` short tokens / single-char CJK.** Gone by construction: `session-search.ts` is now byte-identical to `main` (`git diff upstream/main -- src/store/session-search.ts` is empty). `collectLikeTerms` is untouched, so nothing short-circuits to `{type:'like', terms:[]}`. Your `go`/`ts`/`s3`/`db`/`书` column matches `main` exactly. The dead `else if (explicitOperatorQuery) return []` branch you flagged went out with it.

**2. Mixed-query LIKE net dropping the ASCII half.** Gone: `runLikeSearch` doesn't exist on this branch. `main`'s `runShortCjkFallback` is what runs, unmodified, and it uses the whole trimmed query as one LIKE pattern — there is no per-term set to compute a threshold over. `'名字 kubernetes'` returns `[]` on both sides.

**3. Sub-3-char ASCII tokens discarded.** `isSignificantSearchTerm` is gone — `fts-query.ts` is also byte-identical to `main` now, so the 293-occurrence quote churn went with it and `git blame` there is intact.

I do want to flag one thing honestly rather than claim a clean bill of health. The *set* for `'s3 bucket'` is the same as `main`, but the *order* still differs, and it's the order you called the headline risk:

```
MAIN   ['artifacts go to the s3 bucket named releases', 'the local bucket cache lives in /tmp/bucket']
BRANCH ['the local bucket cache lives in /tmp/bucket', 'artifacts go to the s3 bucket named releases']
```

That is BM25 alone, not the term gate. Cause: since the trigram migration (fdc76a7), `MATCH "s3"` matches **zero rows** — the trigram tokenizer doesn't index a 2-character token at all. So `'s3 bucket'` is effectively a one-term `bucket` search on `main` too, and BM25 then correctly ranks `/tmp/bucket` first because it contains `bucket` twice in a shorter document. Exact scores: `-1.582e-6` vs `-1.227e-6`.

So the discriminating term is invisible to the index before ranking ever runs, on both branches. Your suggested fix (exempt tokens containing a digit) wouldn't reach it either, since it's tokenizer-level, not filter-level. **That's the one open question I can't settle for you:** whether ranking should be relevance-first when the query's most selective term is unindexable. Options as I see them are a `LIKE`-based rescue for sub-3-char tokens, or a query-length-aware tie-break. Both are product calls about the precision/recall trade you'd own, and both are wider than this PR, so I've left them out. Happy to prototype whichever you prefer.

**4. `minTermHitRatio(2) === 1` killing the two-term OR fallback.** Gone: `passesTermHitFilter` and `minTermHitRatio` are no longer imported or defined here. Verified against `main`:

```
'dark chocolate'    both ['user prefers dark mode UI theme']
'docker swarm'      both ['deploy uses docker compose']
'theme preferences' both ['user prefers dark mode UI theme']
```

You were right that the existing three-term test (`name identity Naruto`) doesn't cover this, so I added `still broadens a two-term natural-language query through the OR fallback` pinning `'dark chocolate'`.

## Tests

Three, all in `tests/store/sqlite-memory-store.test.ts`:

- `ranks a dense old match above a fresher, more diluted one (BM25 before recency)` — this is the one that pins the actual change. An old dense row vs. a fresher long diluted one; **it fails on `main` and passes here**, which is also the assertion you asked for in point 3 of your 2026-08-14 comment.
- `still broadens a two-term natural-language query through the OR fallback` — case 4 above.
- `recovers a natural-language query whose raw FTS5 form fails to parse` — worth a correction: **the parse-error catch is already on `main`**, it shipped with #127 (181232b), so this PR no longer adds it. It was in the diff only because the branch predated that merge. The test is a regression guard so the JOIN rewrite doesn't disturb the recovery path — which also answers your point 4 about the retry path, since `runSearch` is the single function both attempts go through, so BM25 applies identically however the query was spelled.

The new tab-indented `tests/store/fts-query.test.ts` is gone entirely — it only tested removed exports.

## Conflicts

None left. Rather than resolve the `CHANGELOG.md` conflict I rebuilt the branch directly on `671a27f` and reapplied only the ranking change, so there was nothing to resolve. The 0.8.x cosmetic reformatting is gone; the CHANGELOG diff is now `+2` — one `## [Unreleased]` bullet under the existing `### Fixed`, scoped to ranking only.

## What moved out

Stopword list (incl. the domain nouns `context`/`question`/`search`/`memory` and the `"of"`/`"一些"` duplicates), `isSignificantSearchTerm`, `collectSignificantSearchTerms`, `passesTermHitFilter`/`minTermHitRatio`, the `runLikeSearch` non-ASCII filtering, all `session-search.ts` changes, and the quote churn.

Two notes on your "also worth fixing" list: the **duplicate LIKE escaping** and the **`fetchLimit` cap on the caller's limit** both lived only in the removed half. `escapeLikePattern` is back to being the single copy in the file, and `fetchLimit` no longer exists, so there's nothing left to fix rather than me having fixed it. Your two latent items about the LIKE tier ordering `last_referenced DESC` and filtering after also apply only to `main`'s `runShortCjkFallback`, which I've left untouched.

On the follow-up PR: I'd planned to open it against #174, but that's closed as completed — the trigram migration solved the CJK-segmentation problem it described, which also makes a good part of the original CJK motivation here moot. So I'd rather not reopen that thread. If you want the recall work pursued, I'm happy to open a fresh issue framing it as tokenizer-and-precision semantics under trigram, using your four cases plus the `'s3 bucket'` ordering finding above as the test set — but only if you think it's worth having, since the ground genuinely moved under it. Your call, and I'm equally fine if you'd rather drop it.

## Verification

Local, on the pushed head `cc847f5`:

- `npm run check` (`ensure-dev` + `tsc --noEmit`) — exit 0, clean
- `npm test` — **all 50 test files passed**
- `tests/store` in detail — **403 passing, 0 failing** across 16 files, `sqlite-memory-store.test.ts` at 48/48 (45 before my three)
- the BM25 test confirmed failing on a clean `upstream/main` worktree (47/48) and passing here (48/48)

CI on this head is green on all four: Lint, Type Check, Tests, Min Supported SDK. PR reports `MERGEABLE` / `CLEAN`.

@chandra447 — this should be the version you described: BM25 in the shape you verified, plus tests, plus one changelog line. Whenever you have a spare cycle, could you take another look? No rush at all. And the offer stands from your side too — if it's faster for you to land the ranking yourself now that it's this small, go ahead, no attribution needed.

chandra447 (commented):
Adversarial review (4 reviewers, unanimous) — no findings. BM25-then-recency ordering is correct, result sets unchanged, edge cases (two-term OR fallback, unparseable NL recovery) pinned by tests. Ship it.

Issue #117, opened 2026-07-21
v0.8.2: compiled Pi fails to load sqlite-native.ts with ResolveMessage for better-sqlite3


atimofeev:
## Summary

`pi-hermes-memory@0.8.2` fails to load under the compiled Pi binary because `src/store/sqlite-native.ts` attempts to resolve `better-sqlite3` with a bare package lookup.

This looks like a regression of the same class as #63, but now introduced through the new shared loader from #114 / v0.8.2.

## Error

```text
[Extension issues]
  ~/.pi/agent/npm/node_modules/pi-hermes-memory/src/index.ts
    Failed to load extension: ResolveMessage: Cannot find package 'better-sqlite3' from
'/home/atimofeev/repos/pi-config/pi-agent/npm/node_modules/pi-hermes-memory/src/store/sqlite-native.ts'
```

`better-sqlite3` is installed and Node can resolve/load it from the extension install root:

```text
/home/atimofeev/repos/pi-config/pi-agent/npm/node_modules/better-sqlite3/lib/index.js
node better-sqlite3 ok
```

So this is not a missing dependency or native ABI mismatch. It is the compiled Pi/Bun resolver failing package resolution from `sqlite-native.ts` before the helper can proceed.

## Environment

```text
Pi: 0.80.7
Install: NixOS packaged Pi compiled binary
Pi binary: ELF x86_64 compiled binary, contains Bun 1.3.13 strings
Node on PATH: v26.5.0
npm: 11.17.0
OS: NixOS Linux x86_64
pi-hermes-memory: 0.8.2
better-sqlite3 dependency: ^12.9.0
```

## Related issues / PRs

- #63 fixed `ResolveMessage: Cannot find package 'better-sqlite3'` by avoiding the Bun path and using `bun:sqlite` under Bun.
- #111 / #114 fixed ABI mismatch by adding the shared `sqlite-native.ts` loader.
- The current failure happens in that new `sqlite-native.ts` loader, so v0.8.2 appears to reintroduce a compiled Pi resolver failure for `better-sqlite3`.

## Workaround tested

A local fallback in `sqlite-native.ts` fixes startup by resolving the installed package via an absolute ancestor `node_modules` path when the bare package lookup fails:

```ts
function resolveBundledBetterSqlite3Entry(): string | null {
  const here = typeof import.meta.url === "string" && import.meta.url.startsWith("file:")
    ? path.dirname(fileURLToPath(import.meta.url))
    : process.cwd();
[...]

PR #124, opened 2026-07-25
perf: cache SQLite connections in AtomicLockCoordinator

Fraguinha:
## Problem

`AtomicLockCoordinator` opens and closes a fresh SQLite connection for every `tryAcquire`, `isCurrentOwner`, and `release` call. Each open runs PRAGMA setup + `CREATE TABLE IF NOT EXISTS`; each close triggers WAL checkpointing. During `session_start`, `syncMarkdownMemoriesToSqlite` calls `withMarkdownMutationLock` per file, and each call also creates a **new** `AtomicLockCoordinator` instance for the same `dbPath`.

### CPU profile (v0.8.2, 2.5 MB sessions.db)

| | Original | With fix |
|---|---|---|
| **better-sqlite3 CPU** | **16.7 s** | **3.7 s** |
| lock-coordinator CPU | 1.2 s | 0.7 s |

This connection churn dominated every `/new` and `/resume`.

## Changes

1. **`atomic-lock-coordinator.ts`** — Add a `_cachedDb` field. `open()` returns the cached handle on subsequent calls instead of creating a new connection each time. Schema setup (PRAGMA + CREATE TABLE) runs once per instance.

2. **`markdown-mutation-lock.ts`** — Cache `AtomicLockCoordinator` instances by `dbPath` in a module-level `Map`, so the same coordinator (and its cached connection) is reused across `withMarkdownMutationLock` calls.

All 420 existing tests pass.


chandra447 (changes requested):
Thanks @Fraguinha — the profile is damning and the diagnosis is right. 16.7s → 3.7s of better-sqlite3 CPU on every `/new` and `/resume` is a big win, and connection churn in `AtomicLockCoordinator` is absolutely the right thing to kill. I want this in. Two things to fix first, one of which is a real leak the profile wouldn't have surfaced.

### 1. Blocking — connections now leak at the three call sites you didn't cache

`coordinatorCache` only covers `markdown-mutation-lock.ts`. There are four `new AtomicLockCoordinator(...)` sites:

```
src/store/markdown-mutation-lock.ts:15   ← cached by this PR
src/handlers/auto-consolidate.ts:64      ← per consolidation attempt
src/store/db.ts:368                      ← per corruption recovery
src/extension-root-migration.ts:229      ← once per migration
```

Before this PR every coordinator was stateless: `open()` … `finally { db.close() }`. Now each instance pins a WAL connection for the life of the process, and `AtomicLockCoordinator` has no `close()`. `tryAcquireConsolidationLock` constructs a fresh coordinator on *every* consolidation attempt, so a long session accumulates open handles (plus their `-wal`/`-shm`) without bound. That's a regression we'd be shipping to every user.

Preferred fix — move the cache into `atomic-lock-coordinator.ts` and route the default-options callers through it, so all four sites get the perf win and nobody has to remember to close:

```ts
const shared = new Map<string, AtomicLockCoordinator>();

/** Process-wide coordinator for `dbPath`; the cached SQLite handle lives as long as the process. */
static shared(dbPath: string): AtomicLockCoordinator {
  const key = path.resolve(dbPath);
  let coordinator = shared.get(key);
  if (!coordinator) {
    coordinator = new AtomicLockCoordinator(dbPath);
    shared.set(key, coordinator);
  }
  return coordinator;
}
```

Keep the public constructor for the option-carrying callers (tests pass `pid` / `incarnation` / `probeIncarnation`, which a `dbPath`-keyed cache would silently ignore).

### 2. Blocking — a failed `ROLLBACK` poisons the cached handle permanently

```ts
} catch (error) {
  try { db.exec('ROLLBACK'); } catch {}
  throw error;
}
```

If `COMMIT` throws and the `ROLLBACK` also throws, `_cachedDb` is left holding an open transaction. The next `tryAcquire` hits `db.exec('BEGIN IMMEDIATE')` — which you moved *outside* the `try` — and SQLite raises `cannot start a transaction within a transaction`, uncaught, forever. The old `finally { db.close() }` rolled back implicitly and handed the next call a fresh connection, so this failure mode didn't exist.

Rare, but silently permanent, and it takes out every markdown mutation in the process. Discard the handle instead of swallowing:

```ts
} catch (error) {
  try {
    db.exec('ROLLBACK');
  } catch {
    try { db.close(); } catch {}
    this._cachedDb = null;
  }
  throw error;
}
```

### Non-blocking

- `_cachedDb` — the codebase uses unprefixed private fields (`private db: DatabaseLike | null = null` in `db.ts`). Worth matching.
- Please add a test for the poisoned-transaction path: force `COMMIT` and `ROLLBACK` to throw on a coordinator, then assert the *next* `tryAcquire` still succeeds. `tests/store/atomic-lock-coordinator.test.ts` already monkey-patches the prototype for the release-retry test, so the seam exists.

I confirmed the interaction you'd expect me to worry about is fine: `DatabaseManager` recovery only renames `sessions.db{,-wal,-shm}` and never touches `.pi-hermes-locks.sqlite`, so cached handles can't end up pointing at a deleted inode.

Happy to push both fixes onto your branch if you'd rather not — just say the word. Otherwise it's yours; I'll re-run CI as soon as you push (the runs were blocked on first-time-contributor approval, not on anything you did — I've approved them and Type Check + Lint are green).

chandra447:
@Fraguinha — I took you up on my own offer and pushed both fixes to your branch rather than leaving you blocked on my review. Your commit is untouched and still authored by you; mine sits on top as a separate commit you can inspect or drop.

**Heads up: I had to rebase, so this was a force-push.** Not a rewrite of your work — `git pull --rebase` (or a fresh `gh pr checkout`) will pick it up cleanly. The reason is #128, which landed after you opened this: it made `atomic-lock-coordinator.ts` resolve its SQLite constructor lazily (`new Database(...)` → `new (getDatabaseCtor())(...)`) because compiled Pi can't resolve `better-sqlite3` at all. That collided with your `open()` change, so a rebase was needed either way and it seemed unkind to hand it to you.

**What I changed, both as requested in the review:**

1. **`AtomicLockCoordinator.shared(dbPath)`** — one coordinator per resolved database path, used by all four call sites (`markdown-mutation-lock`, `auto-consolidate`, `db.ts` recovery, `extension-root-migration`). Your module-level `coordinatorCache` was the right idea; I moved it into the coordinator so the three sites you didn't touch can't silently leak a WAL connection each. `tryAcquireConsolidationLock` was the sharp one — it built a fresh coordinator per consolidation attempt, so a long session would have accumulated open handles without bound. The option-carrying constructor stays public, since a `dbPath`-keyed cache would silently ignore the synthetic `pid`/`incarnation` the tests pass.

2. **Discard the cached handle when `ROLLBACK` fails**, so a failed `COMMIT` can't leave an open transaction on a connection the next caller inherits.

Also renamed `_cachedDb` → `cachedDb` to match the surrounding style.

**Tests added:**

- `recovers when a failed COMMIT leaves the cached connection mid-transaction` — primes the cached handle, breaks both `COMMIT` and `ROLLBACK` on it, asserts the *next* `tryAcquire` still succeeds. I checked it actually fails without the fix: with the discard removed it fails with `cannot start a transaction within a transaction`, exactly the permanent-wedge path.
- `reuses one coordinator per database path so connections are not leaked` — including that a non-normalised path (`dir/./locks.sqlite`) resolves to the same instance.

`npm run check` clean, `npm test` 40/40 files.

**Credit where it's due:** the profile in your description is what made this obvious. 16.7s → 3.7s of better-sqlite3 CPU on every `/new` and `/resume` is the largest single win in the tracker right now, and connection churn in the lock coordinator was not on my radar at all. I've written it up in the changelog under your name.

Give it a look. If you'd rather implement either fix differently, say so and I'll drop my commit — it's your PR. Otherwise I'll merge once you're happy.

chandra447:
Merged — thanks @Fraguinha. This is the biggest single performance win in the tracker and I'd never have found it; the CPU profile in your description did all the work.

For the record, since the branch moved twice under you: I rebased again after #130 and #131 landed (both touched the same changelog section), re-ran the suite, and merged on green. Your commit is preserved in the squash and you're credited in the changelog.

Ships in the next release alongside the #117 compiled-Pi load fix.

PR #127, opened 2026-07-26
fix: retry natural-language queries when raw FTS5 parse fails

### LINKED ISSUE #122: BUG: memory_search silently misses natural-language queries containing uppercase NOT and punctuation
## Summary

`memory_search` can return no results for a natural-language query even when matching memory exists. Query `DO NOT USE FIND /` is classified as raw FTS5 syntax because it contains uppercase `NOT`; unquoted `/` then causes FTS5 parse error. Error is converted to empty result, and safe fallback is skipped.

Same query in lowercase succeeds.

## Environment

```text
Pi: 0.81.1
pi-hermes-memory: 0.8.2
OS: macOS 15.7.7
```

Repository revision reviewed: `4ae04830daa59d90e6b1b99af2cb989090824f47`.

## Reproduction

Store user memory containing:

```text
Never search whole filesystem from root. Do not run find /.
```

Run:

```text
memory_search(query: "DO NOT USE FIND /", target: "user")
```

Actual result:

```text
No memories found matching "DO NOT USE FIND /". Try a different search term or broader query.
```

Control queries against same store:

```text
memory_search(query: "find", target: "user")
```

returns matching entry.

```text
memory_search(query: "do not use find /", target: "user")
```

also returns matching entry.

Direct SQLite control shows raw uppercase query fails:

```text
sqlite> SELECT rowid FROM memory_fts WHERE memory_fts MATCH 'DO NOT USE FIND /';
Error: stepping, fts5: syntax error near "/"
```

## Root cause

Operator detection treats any uppercase `OR`, `AND`, `NOT`, or `NEAR` token as intentional raw FTS5 syntax:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/store/fts-query.ts#L1-L7

`normalizeFts5Query()` then passes whole query through unchanged:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/store/fts-query.ts#L26-L43

`runSearch()` catches FTS5 parse error and returns empty array:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/store/sqlite-memory-store.ts#L736-L756

Fallback is not attempted because `buildFallbackFts5Query()` returns `null` for anything classified as explicit operator query:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/store/fts-query.ts#L45-L64

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/store/sqlite-memory-store.ts#L759-L769

Existing test only asserts malformed operator-only query `AND OR NOT` returns empty. It does not cover normal sentence containing uppercase connector plus punctuation:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/tests/store/sqlite-memory-store.test.ts#L414-L424

## Expected behavior

Tool description accepts natural language. Capitalization should not change recall semantics or silently suppress valid results.

If query is invalid FTS syntax, extension should either retry safely or return explicit query error rather than reporting no matching memories.

## Suggested fix

On FTS5 parse error, always retry a safely quoted natural-language OR query, even when initial input looked like explicit operator syntax. Another option is opt-in raw FTS syntax instead of inferring it from uppercase words.

Safe fallback should preserve filters and limits and should quote or discard punctuation-only terms.

## Tests to add

- `DO NOT USE FIND /` retrieves memory containing `find /`.
- Uppercase natural sentences containing `NOT`, `AND`, `OR`, or `NEAR` behave like lowercase equivalents.
- Invalid raw FTS syntax returns explicit error or performs safe fallback, never false no-results.
- Valid intentional operator queries such as `pnpm OR AEST` keep existing behavior.

cat0825:
Fixes #122

## Problem

`memory_search` (and session search) silently return nothing for natural-language queries containing uppercase operator words plus punctuation, e.g. `DO NOT USE FIND /`:

1. `hasExplicitFts5Operator` sees uppercase `NOT` and classifies the query as raw FTS5 syntax, passing it through unquoted.
2. The bare `/` makes FTS5 raise a parse error.
3. The error is converted to an empty result, and the natural-language fallback is skipped because the query is 'explicit operator', so the user gets "No memories found" even when matching memory exists.

The same query in lowercase succeeds, which makes the behavior surprising.

## Fix

Both search paths now track FTS5 parse errors and recover by retrying the query as natural language:

- `fts-query.ts`: add `normalizeNaturalLanguageFts5Query` / `buildNaturalLanguageFallbackQuery`, which normalize a query as natural language regardless of operator detection.
- `sqlite-memory-store.ts` `searchMemories`: when the raw operator query fails to parse, retry with quoted terms (implicit AND), then the broader OR fallback.
- `session-search.ts` `searchSessions`: same recovery, ending with the existing LIKE fallback.

**Valid** operator queries that legitimately match nothing keep their exact semantics — recovery only triggers on a parse error, so `foo NOT bar` returning empty still returns empty.

## Testing

- Regression tests in both test files: store `Never search whole filesystem from root. Do not run find /.`, query `DO NOT USE FIND /`, assert it is found (exercises both the AND retry and the OR fallback).
- `npx tsc --noEmit` clean.
- `tests/run-all.sh`: all 39 test files pass.

Note: this touches the same area as #118 (BM25 ranking); whichever lands second will need a trivial rebase — happy to do that on my side.

### DIFF
--- a/src/store/fts-query.ts
+++ b/src/store/fts-query.ts
@@ -63,6 +63,37 @@ export function buildFallbackFts5Query(query: string): string | null {
     .join(' OR ');
 }
 
+function quoteTerms(terms: string[], separator: string): string {
+  return terms.map((term) => `"${term.replace(/"/g, '""')}"`).join(separator);
+}
+
+/**
+ * Normalize a query as natural language even when it contains uppercase
+ * operator words. Queries like "DO NOT USE FIND /" are passed through as raw
+ * FTS5 syntax by normalizeFts5Query; when that raw form fails to parse, this
+ * produces the quoted-term form to retry with.
+ */
+export function normalizeNaturalLanguageFts5Query(query: string): string {
+  const trimmed = query.trim();
+  if (trimmed.length === 0) return '';
+
+  return quoteTerms(collectNaturalLanguageTerms(trimmed), ' ');
+}
+
+/**
+ * Build the broader OR fallback for the same recovery path, ignoring
+ * operator detection.
+ */
+export function buildNaturalLanguageFallbackQuery(query: string): string | null {
+  const trimmed = query.trim();
+  if (trimmed.length === 0) return null;
+
+  const terms = collectNaturalLanguageTerms(trimmed);
+  if (terms.length <= 1) return null;
+
+  return quoteTerms(terms, ' OR ');
+}
+
 export function isFts5QueryError(err: unknown): boolean {
   if (!(err instanceof Error)) return false;
   const msg = err.message.toLowerCase();
--- a/src/store/session-search.ts
+++ b/src/store/session-search.ts
@@ -1,5 +1,12 @@
 import { DatabaseManager } from './db.js';
-import { buildFallbackFts5Query, hasExplicitFts5Operator, isFts5QueryError, normalizeFts5Query } from './fts-query.js';
+import {
+  buildFallbackFts5Query,
+  buildNaturalLanguageFallbackQuery,
+  hasExplicitFts5Operator,
+  isFts5QueryError,
+  normalizeFts5Query,
+  normalizeNaturalLanguageFts5Query,
+} from './fts-query.js';
 
 /**
  * Search result from session history.
@@ -93,6 +100,8 @@ export function searchSessions(
   const db = dbManager.getDb();
   const { limit = 10, project, role, since } = options;
 
+  let ftsParseError = false;
+
   const executeSearch = (match: SearchMatch): SessionSearchResult[] => {
     const conditions: string[] = [];
     const params: unknown[] = [];
@@ -160,6 +169,7 @@ export function searchSessions(
       return mapRows(rows);
     } catch (err) {
       if (match.type === 'fts' && isFts5QueryError(err)) {
+        ftsParseError = true;
         return [];
       }
       throw err;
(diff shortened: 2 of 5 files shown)


chandra447 (approved):
Thanks @cat0825 — this is exactly the right shape for #122: recovery keyed on the *parse error* rather than loosening the operator classifier, so intentional FTS5 syntax keeps its semantics. Verified independently against a live store:

| query | result |
|---|---|
| `DO NOT USE FIND /` (target `user`) | finds `Do not run find /.` |
| `do not use find /` | same result — capitalization no longer changes recall |
| `NEVER RUN FIND / AT ROOT!` | recovers too |
| `docker AND kubernetes` | valid operator query, unchanged |
| `pnpm OR AEST` | 2 hits, unchanged |
| `pnpm AND nonexistent` | 0, unchanged |
| `AND OR NOT` | 0, unchanged |
| `DO NOT USE FIND /` + `limit: 1` | limit respected |
| `DO NOT USE FIND /` + `target: memory` | 0 — no filter leakage on the retry path |

The last two were my main worry (the issue explicitly asked that the fallback preserve filters and limits) and both hold, since the retry goes back through `runSearch` with the same conditions.

Merging. Two follow-ups, neither blocking:

1. `normalizeFts5Query` and `buildFallbackFts5Query` still inline the same quote-escaping your new `quoteTerms` helper does — worth collapsing when you rebase.
2. #118 touches all three of these files. Since you own both, your offer to rebase #118 on top of this is the plan — thanks for flagging it up front.

PR #140, opened 2026-07-31
fix: readme.md installation instructions

xz-dev:


chandra447 (approved):
Correct, and it's a real bug rather than a style preference. Thank you — this is the kind of one-liner that quietly costs every new user their first five minutes.

I traced it through Pi's own installer rather than eyeballing it. `parseGitUrl` (`@earendil-works/pi-coding-agent/dist/utils/git.js`) hands its `repo` field straight to `git clone` (`core/package-manager.js:1480`: `await this.runCommand("git", ["clone", source.repo, targetDir])`). Running the three forms through it:

```
git:github:chandra447/pi-hermes-memory      => repo=https://github:chandra447/pi-hermes-memory   <-- current README
git:github.com/chandra447/pi-hermes-memory  => repo=https://github.com/chandra447/pi-hermes-memory <-- this PR
github:chandra447/pi-hermes-memory          => NULL (rejected outright)
```

The current README form *parses* — which is why nobody caught it — but `hosted-git-info` resolves `host`/`path` correctly while leaving `repo` malformed: `github` becomes the hostname and `chandra447` becomes the port. It fails at clone time, not parse time:

```
$ git ls-remote "https://github:chandra447/pi-hermes-memory"
fatal: unable to access '...': URL rejected: Port number was not a decimal number between 0 and 65535

$ git ls-remote "https://github.com/chandra447/pi-hermes-memory"
5aafe2ca04cb...	HEAD
```

Approving as-is.

**Two more instances of the same bug that this PR doesn't cover** — my fault, not yours, and I'll fix them separately rather than expand your PR:

- `AGENTS.md:73` — `pi install github:chandra447/pi-hermes-memory`
- `docs/PUBLISHING.md:105` — same form

Both use the bare `github:` shorthand, which `parseGitUrl` rejects outright (no `git:` prefix means it only accepts explicit protocol URLs — see the docstring on that function). So they're worse than the README form: they never even reach the clone.

Unrelated to this PR, but flagging since your name is on the other two open ones: **CI has never run on any of your branches** — GitHub reports "no checks reported" for #140, #141 and #142. That's a first-time-contributor workflow-approval gate on my side, not anything you did. I'll clear it.

chandra447:
Merged — thank you. 🙏

Great catch, and genuinely a valuable contribution to the extension. This is the kind of fix that's easy to overlook precisely because it *looks* fine: the old spec parsed without error, so nothing complained until you actually tried to clone. Anyone installing from git off the README has been hitting a confusing `Port number was not a decimal number` failure, and you found and fixed it in one line.

For the record, since it took some digging to confirm you were right — Pi's `parseGitUrl` hands its `repo` field straight to `git clone`, and the old form resolved `github` as the hostname and `chandra447` as the port. Your form is the one that produces a valid URL.

Your report also had a knock-on: it made me audit the other install commands in the repo, and two more were broken in a worse way (`AGENTS.md` and `docs/PUBLISHING.md` used a bare `github:` shorthand that `parseGitUrl` rejects outright). Those are fixed in #148, which landed just before this. So one line from you fixed three places.

Sorry about the CI approval gate leaving all three of your PRs without any signal for four days — that was on my side, and it's cleared now. I've left detailed reviews on #141 and #142; both have real blockers, but the underlying diagnoses in each are sound, and in #142's case I had to walk back one of my own criticisms after checking it. Worth a read when you get a chance.

Thanks again for taking the time on this one.

Issue #121, opened 2026-07-23
Policy-only mode makes standing user constraints nondeterministic


jagaliano:
## Summary

A strict standing instruction can be persisted correctly in `USER.md`, `MEMORY.md`, `failures.md`, and SQLite, yet remain absent from model context in later sessions. With default `policy-only` mode, recall depends on model deciding to call `memory_search` before acting. This is unreliable for hard user constraints such as prohibiting unbounded filesystem searches.

This report is a design/reliability gap, not a claim that persistence is broken. Current implementation and documentation intentionally keep stored Markdown memories out of prompt in `policy-only` mode.

## Environment

```text
Pi: 0.81.1
pi-hermes-memory: 0.8.2
OS: macOS 15.7.7
memoryMode: policy-only
memoryPolicyStyle: full
```

Repository revision reviewed: `4ae04830daa59d90e6b1b99af2cb989090824f47`.

## Observed behavior

I repeatedly instructed agent never to run `find /` or other root-wide filesystem searches. Extension successfully persisted this constraint in multiple stores. Despite that, agent later attempted another root-wide search without first calling `memory_search`.

Observed persisted records included:

- `USER.md`: never search whole filesystem from root; scope searches to known directories.
- `MEMORY.md`: global filesystem searches such as `find /` are forbidden.
- `failures.md`: repeated prior violations and corrections.
- `sessions.db` `memories` table: matching `user`, `memory`, and `failure` rows.

Persistence therefore worked. Recall did not happen before action.

## Why this happens

`buildPromptContext()` returns only generic memory policy in `policy-only` mode and never calls `store.formatForSystemPrompt()`:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/src/prompt-context.ts#L25-L42

Tests explicitly require policy-only context to exclude stored memory blocks:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/tests/handlers/prompt-context.test.ts#L5-L40

README documents same behavior:

https://github.com/chandra447/pi-hermes-memory/blob/4ae04830daa59d90e6b1b99af2cb989090824f47/README.md#L151-L164

Generic memory policy tells model when to search, but extension cannot guarantee model performs search before every relevant tool call. A persisted standing instruction is therefore probabilistic rather than continuously active.

## Expected behavior

[...]

PR #141, opened 2026-07-31
feat(ui): add expandable tool result views

xz-dev:
## Summary

- add one shared collapsed/expanded TUI result view for the four logical Hermes tools: `memory`, `memory_search`, both `session_search` variants, and `skill_manage`
- follow Pi's existing global tool-output expansion state (`app.tools.expand`; Ctrl-O by default) without registering a shortcut or owning extension-local expansion state
- keep the collapsed view to one actionable line while expanded mode shows the complete already-bounded tool result
- sanitize only the display copy before it reaches the terminal while preserving model-visible `content` and `details` exactly

## Motivation

Without custom `renderResult` functions, Pi's generic fallback can print complete custom-tool results even when global tool output is collapsed. Search and full-document results can therefore dominate the transcript.

A custom renderer also replaces Pi's generic display boundary. This implementation therefore strips terminal control sequences from its display-only copy before deriving summaries or creating TUI `Text`; the original tool result remains unchanged for the model and extension logic.

This PR remains limited to tool-result presentation. Slash-command reports keep their existing notification/UI behavior. The change does not add custom transcript entries, alter command handlers, raise the minimum Pi SDK, or change source-level output safety limits.

## Behavior

- **Collapsed:** one width-safe summary line with Pi's configured expansion key hint
- **Expanded:** the complete safe display copy of the text already returned by the tool, still subject to each tool's existing source-level limits
- **Terminal safety:** OSC, CSI/SGR styling, carriage returns, and binary controls cannot reach rendered rows; `strip-ansi` is applied only in the renderer
- **Failures:** actionable reasons remain visible through error foreground text
- **Shell state:** pending/error/success card background restoration follows Pi's actual partial and `context.isError` state, not application-level `details.success`
- **Warning-bearing successes:** warning text remains visible instead of being hidden behind a generic success label
- **Failure memories:** successful saves render as `Saved · target: failure · category: …`, avoiding a misleading `failure:` prefix

Tool-specific summaries cover memory actions and targets, search/session counts, and skill outcomes. Structured skill failures such as `{ "success": false, "error": "..." }` are rendered as failures rather than successful JSON output.

## Compatibility and scope

- keeps the existing project minimum Pi SDK (`>=0.80.1`)
- uses public `renderResult`, `ToolRenderContext.isError`, `keyHint("app.tools.expand", ...)`, and TUI APIs available at that floor
- adds the pinned runtime dependency `strip-ansi@7.2.0` because Pi's internal fallback sanitizer is not exported from the public package API in any release through `0.83.0`
- no slash-command or command-handler changes
- no `appendEntry` / `registerEntryRenderer` usage
- no extension factory/index wiring changes
- no model-context, persistence, or tool execution changes

## Verification

- focused renderer and wiring tests: **13/13 passed**
- `npm run check`: passed
- `npm run check:min-sdk`: passed against Pi SDK **0.80.1**
- `npm test`: all **44 test files passed**
- production dependency audit: **0 vulnerabilities**
- `npm pack --dry-run`: passed
- `git diff --check`: clean
- independent review: **approved**, no remaining findings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,8 @@
       "license": "MIT",
       "dependencies": {
         "@earendil-works/pi-tui": "^0.80.2",
-        "better-sqlite3": "^12.9.0"
+        "better-sqlite3": "^12.9.0",
+        "strip-ansi": "7.2.0"
       },
       "devDependencies": {
         "@earendil-works/pi-ai": "^0.80.2",
@@ -3269,6 +3270,18 @@
         "node": ">= 14"
       }
     },
+    "node_modules/ansi-regex": {
+      "version": "6.2.2",
+      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.2.2.tgz",
+      "integrity": "sha512-Bq3SmSpyFHaWjPk8If9yc6svM8c56dB5BAtW4Qbw5jHTwwXXcTLoRMkpDJp6VL0XzlWaCHTXrkFURMYmD0sLqg==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=12"
+      },
+      "funding": {
+        "url": "https://github.com/chalk/ansi-regex?sponsor=1"
+      }
+    },
     "node_modules/base64-js": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
@@ -4109,6 +4122,21 @@
       "license": "MIT",
       "dependencies": {
         "safe-buffer": "~5.2.0"
+      }
+    },
+    "node_modules/strip-ansi": {
+      "version": "7.2.0",
+      "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.2.0.tgz",
+      "integrity": "sha512-yDPMNjp4WyfYBkHnjIRLfca1i6KMyGCtsVgoKe/z1+6vukgaENdgGBZt+ZmKPc4gavvEZ5OgHfHdrazhgNyG7w==",
+      "license": "MIT",
+      "dependencies": {
+        "ansi-regex": "^6.2.2"
+      },
+      "engines": {
+        "node": ">=12"
+      },
+      "funding": {
+        "url": "https://github.com/chalk/strip-ansi?sponsor=1"
       }
     },
     "node_modules/strip-json-comments": {
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
   },
   "dependencies": {
     "@earendil-works/pi-tui": "^0.80.2",
-    "better-sqlite3": "^12.9.0"
+    "better-sqlite3": "^12.9.0",
+    "strip-ansi": "7.2.0"
   }
 }
--- a/src/tools/memory-search-tool.ts
+++ b/src/tools/memory-search-tool.ts
@@ -4,6 +4,8 @@ import { StringEnum } from "@earendil-works/pi-ai";
 import { DatabaseManager } from '../store/db.js';
 import { searchMemories, getMemoryStats } from '../store/sqlite-memory-store.js';
 import type { MemoryCategory } from '../types.js';
+import { createSharedToolResultRenderer } from './shared-output-view.js';
+import { searchResultView } from './tool-result-views.js';
 
 interface SearchResult {
   success: boolean;
@@ -31,6 +33,7 @@ Returns matching memory entries with project context and dates.`,
       'Use memory_search to find project-specific memories or user preferences.',
       'Use memory_search with category filter to find specific types of memories (failure, correction, insight, etc.).',
     ],
+    renderResult: createSharedToolResultRenderer(searchResultView),
     parameters: Type.Object({
       query: Type.String({ description: 'Search query. Use natural language or specific terms.' }),
       project: Type.Optional(Type.String({ description: 'Filter by project name. Pass null for global memories only.' })),
--- a/src/tools/memory-tool.ts
+++ b/src/tools/memory-tool.ts
@@ -20,6 +20,8 @@ import {
 } from "../store/sqlite-memory-store.js";
 import { MEMORY_TOOL_DESCRIPTION } from "../constants.js";
 import type { MemoryCategory, MemoryResult } from "../types.js";
+import { createSharedToolResultRenderer } from "./shared-output-view.js";
+import { memoryResultView } from "./tool-result-views.js";
 
 function appendSyncWarning(result: MemoryResult, warning: string): MemoryResult {
   const warnings = [...(((result as any).warnings ?? []) as string[]), warning];
@@ -241,6 +243,7 @@ export function registerMemoryTool(
       "Do NOT use memory for temporary task state, TODO items, or session progress — only for durable, cross-session facts.",
       "Use target='failure' with category to save what didn't work (failures, corrections, insights).",
     ],
+    renderResult: createSharedToolResultRenderer(memoryResultView),
     parameters: Type.Object({
       action: StringEnum(["add", "replace", "remove"] as const),
       target: StringEnum(["memory", "user", "project", "failure"] as const),
--- a/src/tools/session-search-tool.ts
+++ b/src/tools/session-search-tool.ts
@@ -8,6 +8,8 @@ import { searchSessionAnchors } from '../store/session-anchor-search.js';
 import type { SessionAnchorRange, SessionAnchorSearchResult } from '../store/session-anchor-search.js';
 import type { SessionSearchConfig } from '../types.js';
 import { AGENT_ROOT } from '../paths.js';
+import { createSharedToolResultRenderer } from './shared-output-view.js';
+import { searchResultView } from './tool-result-views.js';
 
 interface SearchResult {
   success: boolean;
@@ -90,6 +92,7 @@ exclude:
       'Request source anchors, not summaries or previews.',
       'Use all for required terms, any for alternatives, and exclude for terms that must not appear in a returned range.',
     ],
+    renderResult: createSharedToolResultRenderer(searchResultView),
     parameters: Type.Object({
       markdown: Type.String({ description: 'Markdown request with optional from/to/cwd/limit fields and all/any/exclude lists.' }),
     }),
@@ -157,6 +160,7 @@ Returns bounded conversation snippets with session dates and project context. La
       'Use session_search when the user asks about previous discussions or past work.',
       'Use session_search when you need context from earlier sessions.',
     ],
+    renderResult: createSharedToolResultRenderer(searchResultView),
     parameters: Type.Object({
       query: Type.String({ description: 'Search query. Use natural language or specific terms.' }),
       project: Type.Optional(Type.String({ description: 'Filter by project name (optional).' })),
--- a/src/tools/shared-output-view.ts
+++ b/src/tools/shared-output-view.ts
@@ -103,7 +103,14 @@ function renderView(
   theme: SharedOutputTheme | undefined,
   background: ToolCardBackground,
 ): Component {
-  if (options.expanded) return new Text(view.expandedText || view.summary, 0, 0);
+  if (options.expanded) {
+    return new Text(
+      view.expandedText || view.summary,
+      0,
+      0,
+      (line) => restoreBackground(line, background, theme),
+    );
+  }
 
   return {
     render(width: number): string[] {
@@ -127,20 +134,6 @@ function renderView(
     },
     invalidate(): void {},
   };
-}
-
-export function renderSharedToolResult(
-  result: unknown,
-  options: ToolRenderResultOptions,
-  theme?: SharedOutputTheme,
-): Component {
-  const view = normalizeSharedOutputView(result);
-  const background: ToolCardBackground = options.isPartial
-    ? "toolPendingBg"
-    : view.status === "failure"
-      ? "toolErrorBg"
-      : "toolSuccessBg";
-  return renderView(view, options, theme, background);
 }
 
 export function createSharedToolResultRenderer(
--- a/src/tools/shared-output-view.ts
+++ b/src/tools/shared-output-view.ts
@@ -1,4 +1,5 @@
 import { keyHint, type ToolRenderResultOptions } from "@earendil-works/pi-coding-agent";
+import stripAnsi from "strip-ansi";
 import {
   Text,
   sliceByColumn,
@@ -38,6 +39,12 @@ function textBlocks(content: unknown): string[] {
   });
 }
 
+function sanitizeDisplayText(text: string): string {
+  return stripAnsi(text).replace(/[\p{Cc}\p{Cs}\uFFF9-\uFFFB]/gu, (character) =>
+    character === "\n" || character === "\t" ? character : ""
+  );
+}
+
 function firstLine(text: string): string {
   return text.split(/\r?\n/).find((line) => line.trim())?.trim() ?? "";
 }
@@ -52,12 +59,13 @@ function reason(details: Record<string, unknown> | null): string {
 export function normalizeSharedOutputView(input: unknown): SharedOutputView {
   const result = record(input);
   const details = record(result?.details);
-  const expandedText = textBlocks(result?.content).join("\n");
+  const expandedText = sanitizeDisplayText(textBlocks(result?.content).join("\n"));
+  const detailsReason = sanitizeDisplayText(reason(details));
   const failure = result?.isError === true || details?.success === false || details?.isError === true;
   const status: SharedStatus = failure ? "failure" : expandedText.trim() ? "success" : "empty";
   const summary = failure
-    ? reason(details) || firstLine(expandedText) || "Error"
-    : firstLine(expandedText) || reason(details) || "No output";
+    ? detailsReason || firstLine(expandedText) || "Error"
+    : firstLine(expandedText) || detailsReason || "No output";
   return { summary, expandedText, status };
 }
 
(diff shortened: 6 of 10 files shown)


xz-dev:
Updated the branch with `d924726` (`fix(ui): clarify memory result summaries`).

Verification on the updated branch:
- `npx tsx --test tests/tools/memory-tool.test.ts` — 36/36 passed
- `npm run check` — passed
- `npm test` — all 45 test files passed (790 tests)
- independent review — approved with no blocking findings

The change is limited to collapsed TUI summary derivation; model-visible content/details and expanded output remain unchanged.

chandra447 (changes requested):
This is a lot of careful work — the architecture is sound, the collapsed/expanded model is the right shape, and I want to call out two things you got right that I specifically went looking for:

- `pi.appendEntry` maps to `appendCustomEntry` (type `"custom"`, TUI-only), **not** `appendCustomMessageEntry` (type `"custom_message"`, documented as participating in LLM context). Command output does not leak into the model's context. That's an easy trap and you avoided it.
- `styleExpandedBody` returning memory content verbatim into `new Text(...)` looked like an ANSI-injection sink on first read. It isn't — Pi's own default tool renderer is the identical component with identical raw text (`dist/modes/interactive/components/tool-execution.js:114`). You match framework behaviour exactly.

That said, I can't merge it as-is, and I want to be straight with you about scope before the specifics.

## The scope question, honestly

This is a 34-file, 4000-line UI feature with no linked issue, in an extension whose entire UI surface is slash commands and one interactive skills view. I didn't ask for it, and the decision to take on a shared rendering layer here is mine to make before you spend more time on it. That's on me for not having a CONTRIBUTING note about pre-agreeing features this size — I'll add one.

I'm not rejecting the idea. But please don't invest further until we've agreed the shape, because right now the review cost is high enough that it's blocking your two smaller PRs from getting attention.

## Blockers

**1. The extension fails to load entirely on Pi < 0.80.5.**

`src/index.ts:192` calls `registerSharedOutputView(pi)` unguarded, and `shared-output-view.ts:417` calls `pi.registerEntryRenderer(...)`. That method doesn't exist before 0.80.5 — confirmed by unpacking the published tarballs (`0.80.3: 0 hits`, `0.80.5: 1 hit`). Against the SDK pinned on main:

```
src/tools/shared-output-view.ts(417,5): error TS2339:
  Property 'registerEntryRenderer' does not exist on type 'ExtensionAPI'.
```

At runtime on such a host the factory throws, and Pi's loader treats a throwing factory as a load failure and drops the whole extension (`loader.js:378` `await factory(api)` → `:395` `errors.push(...); continue;`). The user loses the memory tool, memory injection and every command — not just the new view.

The `peerDependencies` bump to `>=0.80.5` doesn't protect them: this package ships source (`"main": "src/index.ts"`) and is loaded in-process by whatever Pi binary the user is running, so the peer range is advisory with no runtime gate. Everything else you use (`renderResult`, `Theme`, `Text`, `sliceByColumn`, `truncateToWidth`, `visibleWidth`, `appendEntry`) already exists in 0.80.2 — `registerEntryRenderer` is the *single* symbol forcing a six-minor floor bump. Feature-detect it and fall back to `ctx.ui.notify`.

**2. `npm test` fails on macOS, and takes 40 other tests down with it.**

`tests/tools/session-search-tool.test.ts:454` renders at a fixed width of 120 and asserts `/matched any: needle/`. The rendered line embeds an absolute `os.tmpdir()` path; on macOS that's `/var/folders/<hash>/T/`, pushing the anchor to ~123 columns so `Text` hard-wraps between `any:` and `needle`. Proven environment-dependent:

```
TMPDIR=/tmp/t2 npx tsx --test tests/tools/session-search-tool.test.ts   -> pass 12 / fail 0
default macOS TMPDIR                                                    -> pass 11 / fail 1
```

`tests/run-all.sh` uses `set -euo pipefail` and exits on first failure, so this also prevents `shared-output-view.test.ts` (19 tests) and `skill-tool.test.ts` (21 tests) from running at all. Assert against the view model, not the width-120 rendering.

**3. Command output goes silent in every non-TUI runtime.**

`skills-command.ts:1288` is the *explicit headless branch* (`!ctx.hasUI || typeof ctx.ui.custom !== "function"`) — and it now emits a TUI-only custom entry that needs a registered renderer to be visible. `ctx.ui.notify` worked without a UI; `appendEntry` doesn't. Same substitution in eight more handlers: `insights.ts:74`, `preview-context.ts:44` and `:87`, `learn-memory.ts:183`, `switch-project.ts:43` and `:75`, `index-sessions.ts:83`, `sync-markdown-memories.ts:261`, `auto-consolidate.ts:287`.

Related and also worth stating in the PR body: `getEntryRenderer` is only referenced from `interactive-mode.js:2549`, so **RPC clients** that listen for `notify` (`rpc-mode.js:86-93`) stop seeing command output entirely and would have to parse raw `entry_appended` events.

**4. A documented crash guard was removed.**

`auto-consolidate.ts:287` dropped this, added deliberately by `3af7025` ("fix: avoid stale ctx crash in memory-consolidate"):

```ts
try { ctx.ui.notify(summary, "info"); } catch {
  // Child consolidation can indirectly trigger a runtime reload/session
  // replacement. If that happens, the original command ctx is stale ...
}
```

Switching from `ctx` to `pi` doesn't dodge it: `appendCommandReport` → `pi.appendEntry` → `runtime.assertActive()` (`loader.js:242-245`), which throws once the runtime is invalidated (`agent-session.js:573`). The exact scenario the comment describes now produces an unhandled rejection. You kept the two *earlier* progress guards and removed only the final one — the one most likely to be stale.

## Majors

**5. SQLite sync warnings disappear behind a green "Saved".** `primaryMessage` (`tool-result-views.ts:120-122`) splits on `/\s*\bWarning:/` and keeps only the part before it; `data.warning`/`data.warnings` — which `appendSyncWarning` populates (`memory-tool.ts:26-35`) — are never read, and `deriveStatus` still returns `success`. With a real `appendSyncWarning` payload the collapsed row renders `Saved · target: memory · 42 entries · 42/200 (details hidden)`. Collapsed is the default (`interactive-mode.js:206`, applied unconditionally at `:2400`), and `--verbose` doesn't change it. For an extension whose SQLite mirror is a core invariant, hiding "SQLite sync failed" is the one regression here I'd call serious.

**6. Consolidation failure is encoded only in colour.** `summary: \`Memory consolidation: ${results.length} targets\`` with the pass/fail signal passed solely through `status`, which only changes `themeFg(...)`. A run where every target failed renders identical text to a clean run, differing only in colour — gone under `NO_COLOR`. Fold the count in: `1 of 3 targets failed`.

**7. Merge conflict with #147.** Verified by a real test merge: `CONFLICT (content): Merge conflict in src/handlers/preview-context.ts`. Beyond the text, it's semantically incompatible — #141 hardcodes the policy-only count as `policyPrompt ? 1 : 0` while #147 introduces a `blockCount` that also counts an injected standing-instructions block, so a naive resolution reports "1 block" while two are injected. #147 closes #121 and will land first; please rebase after.

## Minors

- **Collapsed row overflows by 13 columns when `isPartial`.** `styleSummaryLine` prepends `"In progress: "` *after* the width budget is spent (`shared-output-view.ts:308-329`). Measured: width 40 → 53 columns, width 80 → 93. Latent today (no tool calls `onUpdate`) but you ship and test `isPartial` as a contract, and the guard test asserts only `collapsedRows.length === 1`, never `visibleWidth(row) <= width`.
- **Use Pi's public `keyHint`** instead of hardcoded `" (details hidden)"`. Pi's own bash renderer does exactly this (`core/tools/bash.js:166-169`), and both `keyHint` and `truncateToVisualLines` are exported from the package root — no deep import needed. Your own test comment at `shared-output-view.test.ts:554` already identifies this as the better option.
- **Drop the test-only options seams.** `registerIndexSessionsCommand`'s `{ sessionsDir, memoryDir, createDatabaseManager }` and `registerSwitchProjectCommand`'s `{ projectsDir }` are unused by production callers and add path-normalization bypasses to exported APIs in the one area where `src/paths.ts` normalization is the security control. Your own new `extension-shared-output-wiring.test.ts:15-17` shows the better pattern: set `PI_CODING_AGENT_DIR` before import.
- **Unrelated changes to split out:** the peer floor bump + lockfile churn (+316/-1719 — 88% of this PR's deletions); `switch-project.ts:65` quietly changing entry counting from `.filter(Boolean)` to `.filter(e => e.trim())`; and `PLAN/shared-expandable-output-view.md`, which looks like your private working plan and isn't gitignored.
- **Indentation:** `shared-output-view.ts` is hard tabs throughout (284 lines); the repo is 2-space, including your own sibling files `tool-result-views.ts` and `shared-command-report.ts` in this same PR.

## Security

I ran a dedicated security pass over this diff, given its size and that this extension sits next to credentials. **It came back clean** — no new exec, no network egress, no new dependencies, no install scripts, no path-normalization bypass in a reachable path, no content-scanner bypass, no eval. Recording that explicitly so it's on the record rather than implied.

Let's agree the feature's shape first. If you're up for it, open an issue describing the end state you want and I'll respond there before you touch more code.

chandra447:
CI is running now — first-time-contributor approval gate on my side, sorry.

**All three jobs pass.** I want to be straight about what that does and doesn't mean, because green here is misleading and I don't want you to read it as the blockers being cleared.

## Why Type Check passing doesn't clear blocker 1

CI runs `npm install`, which honours your bumped `^0.80.5` and resolves **0.80.10** — a version that *has* `registerEntryRenderer`. Confirmed from your lockfile:

```
node_modules/@earendil-works/pi-coding-agent  0.80.10
```

So the job tested the SDK the PR asks for, which was never in question. The blocker is about the SDK a *user already has*. This package ships source (`"main": "src/index.ts"`) and is loaded in-process by whatever Pi binary they run, so `peerDependencies` is advisory with no runtime gate.

I reproduced the actual failure: your branch's source, against the SDK currently pinned on `main` (0.80.2, i.e. what an existing user has), with an `ExtensionAPI` stub exposing exactly the 0.80.2 surface:

```
EXTENSION FACTORY THREW: pi.registerEntryRenderer is not a function
registered before throw: (nothing)
```

`registered before throw: (nothing)` is the part that matters. The throw happens at `src/index.ts:192`, before a single tool or command is registered. And Pi's loader drops the whole extension on a throwing factory (`core/extensions/loader.js:377-381`):

```js
const { extension, error } = await loadExtension(extPath, ...);
if (error) { errors.push({ path: extPath, error }); continue; }
```

So the user loses the memory tool, memory injection, and every slash command — for a rendering feature. Feature-detect `registerEntryRenderer` and fall back to `ctx.ui.notify` and this disappears entirely; it's the one symbol forcing the floor bump.

## Why Tests passing doesn't clear blocker 2

The failing assertion is TMPDIR-length-dependent. CI runners are Linux, where `TMPDIR` is `/tmp` — short enough that the line doesn't wrap. macOS gives `/var/folders/<hash>/T/`, pushing the rendered anchor past the fixed width of 120 and wrapping between `any:` and `needle`:

```
TMPDIR=/tmp/t2 npx tsx --test tests/tools/session-search-tool.test.ts   -> pass 12 / fail 0
default macOS TMPDIR                                                    -> pass 11 / fail 1
```

Since `tests/run-all.sh` uses `set -euo pipefail`, that one failure also stops `shared-output-view.test.ts` (19 tests) and `skill-tool.test.ts` (21 tests) from ever running on a Mac. Asserting against the view model instead of the width-120 rendering fixes it and makes the test better anyway.

## A correction in your favour

I criticised the floor bump `>=0.74.0` → `>=0.80.5` as dropping existing users. That was wrong, and I checked it afterwards: `src/handlers/review-memory-ops.ts` imports `@earendil-works/pi-ai/compat`, which doesn't exist before 0.80.1 (`ERR_PACKAGE_PATH_NOT_EXPORTED`). The real floor was already 0.80.1; `>=0.74.0` was fiction.

So your instinct to raise it was right — it just wasn't verifiable, because nothing checked. #149 corrects the floor and adds a CI job that installs the declared minimum and type-checks against it. Once that lands, `>=0.80.5` becomes a normal reviewable change with CI behind it.

That does **not** dissolve blocker 1, though, and I want to be clear why: a floor of `>=0.80.5` in `package.json` still doesn't stop a user on 0.80.2 from loading this extension and losing everything. npm peer ranges warn; they don't gate an in-process source load. The feature detection is still needed.

The rest of my review stands — headless output, the stale-ctx guard, the hidden SQLite warnings, the #147 conflict. And the scope conversation is still the one I'd like to have first.

xz-dev:
Reworked this PR in place and force-pushed `aa74e4f` to narrow the scope to **tool results only**, based on current `main` (`fe35a63`).

Removed from the PR scope:
- slash-command report rendering and custom transcript entries
- `registerEntryRenderer` / `appendEntry`
- minimum-SDK and dependency/lockfile changes
- command-handler, extension-index, and PLAN changes

The branch now only adds shared `renderResult` behavior for `memory`, `memory_search`, both `session_search` variants, and `skill_manage`. It also incorporates the relevant review fixes: textual warnings/errors, skill `success:false` classification, ANSI/CJK width handling, and tool-card background continuity in both collapsed and expanded views.

Verification: focused 10/10, long-TMPDIR 18/18, `npm run check` passed with locked Pi SDK 0.80.2, all 44 test files passed, pack dry-run passed, and independent review + final merge gate approved.

chandra447 (changes requested):
The force-pushed scope is a substantial improvement. The old compatibility, command-output, stale-context, dependency, and merge-conflict blockers are gone; the current diff is limited to public `renderResult` wiring and model-visible `content`/`details` remain unchanged. I also confirmed it type-checks at the declared 0.80.1 floor and passes the full suite on macOS.

Two renderer issues remain:

### 1. Blocker: expanded results bypass Pi's terminal-output sanitization

Pi's generic custom-tool fallback does not pass raw result text directly to `Text`: `getTextOutput()` runs every text block through `stripAnsi()` and `sanitizeBinaryOutput()` first. `normalizeSharedOutputView()` now copies raw tool text into `expandedText`, and expanded mode passes that directly to `new Text(...)`; `restoreBackground()` only handles SGR background resets.

This is reachable through existing tools, not a hypothetical input: `session_search` returns stored user/assistant snippets verbatim, and `memory_search` returns stored memory content. A stored OSC sequence therefore reaches the terminal when the result is expanded. I reproduced an OSC 52 payload on this head; `component.render()` retained `\u001b]52;...\u0007` byte-for-byte.

Please sanitize a display-only copy before deriving summaries or creating `Text`, matching Pi's prior fallback boundary. Do not mutate the returned/model-visible `content` or `details`. The renderer tests should assert that OSC, non-SGR CSI, ANSI styling, carriage returns, and binary controls cannot reach rendered rows; the current test that preserves raw ANSI should be inverted.

### 2. Major: logical failures use a background different from Pi's actual shell

Pi's outer `ToolExecutionComponent` chooses the card background from the real tool result's `isError`. Application-level failures such as `{ success: false, message: "query is required" }` return normally, so the outer shell remains `toolSuccessBg`. This renderer instead chooses `toolErrorBg` from `details.success` and reapplies it only after SGR sequences.

The result is a split card: success-colored padding/title, then an error-colored collapsed text region; expanded plain text has no SGR to trigger restoration and returns to an entirely success-colored row. I reproduced that collapsed/expanded mismatch.

Within this renderResult-only scope, derive the restoration background from Pi's shell state: partial -> `toolPendingBg`; otherwise `context.isError` -> `toolErrorBg`; otherwise `toolSuccessBg`. Keep logical failures visible through error foreground plus their actionable text.

Validation on head `aa74e4f`: `npm run check`, `npm run check:min-sdk`, and all 44 test files pass. Structured review otherwise found the public SDK contract, warning/error summaries, width handling, and tool wiring sound.

xz-dev:
Added `266a300` to preserve the failure-memory wording clarification from the earlier branch. A successful result such as `Failure memory saved: tool-quirk` now has an explicit regression test for the outcome-first collapsed summary:

```text
Saved · target: failure · category: tool-quirk · 16 entries · 92% — 9271/10000 chars
```

This avoids the misleading old `failure: ...` prefix while preserving the expanded/model-visible JSON unchanged. Focused renderer/wiring tests are 11/11 and `npm run check` passes.

xz-dev:
Addressed review `pullrequestreview-4853650799` in signed commit `8f943bc`:

- sanitize a display-only copy before summary derivation and expanded `Text`; tests cover OSC (BEL/ST), non-SGR CSI, SGR styling, carriage returns, and binary controls
- preserve the original model-visible `content` and `details` unchanged
- derive restored card backgrounds from Pi shell state: partial → pending, `context.isError` → error, otherwise success; logical failures remain actionable/error-foreground text
- retain the successful failure-memory summary (`Saved · target: failure · category: …`)

I initially investigated raising the minimum Pi SDK as the cleanest reuse path. However, no published `@earendil-works/pi-coding-agent` release through `0.83.0` exports `getTextOutput`, `stripAnsi`, or `sanitizeBinaryOutput` from its public package API, so a version bump would not provide a supported sanitizer. Instead this follow-up adds pinned runtime dependency `strip-ansi@7.2.0`; the Pi floor remains truthfully `>=0.80.1`.

Verification on exact head `8f943bc6e778905bec81542d733f02cb82f67aa0`:
- focused renderer/wiring: 13/13
- `npm run check`
- `npm run check:min-sdk` against Pi 0.80.1
- `npm test`: all 44 test files
- production dependency audit: 0 vulnerabilities
- `npm pack --dry-run`
- independent review: APPROVE, no findings

chandra447 (approved):
Reviewed latest head 8f943bc6e778905bec81542d733f02cb82f67aa0. Pi-native renderResult implementation, display-only sanitization, context.isError handling, and public SDK compatibility verified; CI checks are green.

Issue #139, opened 2026-07-31
OpenCodeGo auth key not refreshed after rotation — uses stale cached key


avinashkanaujiya:
### What happened?

The pi-hermes-memory extension caches the OpenCodeGo subscription auth key and never refreshes it. When the pi-opencode-go-rotation extension (`@lnilluv/pi-opencode-go-rotation`) auto-rotates the key (due to hitting weekly usage limits), everything else picks up the new key on the next attempt — but pi-hermes-memory keeps using the stale one.

### Steps to reproduce

1. Use `opencode-go` as the model provider with a subscription key
2. Hit the weekly usage limit so `pi-opencode-go-rotation` rotates to a new key
3. Observe that other extensions / direct API calls work fine with the new key
4. pi-hermes-memory's background review / memory writes fail because it's still using the old key

### Expected behavior

pi-hermes-memory should pick up the current auth key from whichever source pi-opencode-go-rotation uses to store it — or at minimum re-read it on each call rather than caching indefinitely.

### Environment

- pi-hermes-memory version: latest (`npm:pi-hermes-memory`)
- pi-opencode-go-rotation: latest (`@lnilluv/pi-opencode-go-rotation`)

chandra447:
@avinashkanaujiya — real bug, confirmed, and thank you for reporting it. This is a nasty one to notice from the outside: everything else in your setup picks up the rotated key, so the natural conclusion is that pi-hermes-memory is broken generally rather than that it's specifically holding a stale credential. You did the hard part by isolating it to this extension and naming the rotation tool. Thank you for sticking with the extension through it instead of uninstalling.

Traced it through the SDK rather than guessing:

- `AuthStorage` calls `this.reload()` exactly once, in its constructor. For an `api_key` credential `this.data` is never re-read again — the only automatic reload is on an OAuth refresh *failure*, which never fires for a plain key.
- `ExtensionRunner` exposes `modelRegistry` as a getter onto one instance created in its constructor, so every event handler sees the same singleton, and with it the same frozen credentials, for the whole process lifetime.

So the direct in-process path resolved auth against a snapshot taken at session start. When `pi-opencode-go-rotation` writes the new key to `auth.json`, a running session never sees it. Your intuition about the cause was right, and your suggested remedy — *"re-read it on each call rather than caching indefinitely"* — is exactly what the fix does.

[...]

PR #142, opened 2026-07-31
feat(memory): apply consolidation plans atomically

xz-dev:
## Summary

- add `MemoryStore.applyMutationPlan()` for single-target, all-or-nothing memory mutations under the existing cross-process Markdown mutation boundary
- apply direct consolidation operation batches with one atomic plan, rejecting empty, mixed-target, unexpected-target, invalid, over-limit, and non-shrinking plans without publishing partial changes
- preserve failure-memory category, reason, and project attribution, including same-text failures scoped to different projects
- keep the current direct-to-subprocess fallback, provider-auth refresh, notifications, session flush behavior, SQLite reconciliation, and minimum SDK unchanged

## Why this PR was narrowed

The original version of this PR bundled provider-auth refresh, transport-policy changes, dependency updates, and atomic consolidation work. Provider auth was subsequently fixed upstream by #146 using `authStorage.reload()`, and #149 established the actual minimum SDK and minimum-version CI.

This rewrite retains only the independently useful portion identified in review: atomic mutation plans. It also fixes the review finding that `operation.project` was accepted by the store but never populated by `applyReviewOperations`, silently dropping project attribution from failure-memory additions.

### Store primitive

`applyMutationPlan(target, operations, { requireShrink? })`:

- runs within the existing `runTargetMutation()` lock and external-write retry machinery
- synchronizes the target from disk before planning
- evaluates all operations sequentially against a draft, so later operations can use earlier draft results
- scans both add and replacement content
- rejects empty content/lookup text, duplicate additions, missing or ambiguous matches, final over-capacity state, and non-shrinking plans when required
- preserves creation/project metadata on replacement
- formats failure additions with category/reason and stores project attribution
- allows the same failure text in distinct project scopes while rejecting duplicates within one scope
- publishes only after the complete plan validates, using one final `setEntries()`/`saveToDisk()` path
- does not recursively invoke the auto-consolidator

### Consolidation bridge

For explicit atomic consolidation:

- every operation must use one logical target
- the target must match the caller's expected target
- project plans write only to `projectStore`; unavailable project storage rejects the complete plan
- the complete operation array is passed to `applyMutationPlan()` once with strict shrink enforcement
- success applies the entire plan; failure applies none of it
- failure additions without an explicit category default to `failure`, matching ordinary non-atomic behavior

Direct auto-consolidation now requests this atomic/shrinking contract. A failed, empty, or rejected direct result still follows the existing subprocess fallback; this PR does not change transport policy.

## Explicit non-goals

This rewrite does **not** include the old PR's:

- provider-runtime or auth implementation; #146 is retained unchanged
- removal of direct-to-subprocess fallback
- background-review notification changes
- session-flush error propagation changes
- SDK, dependency, peer-floor, or lockfile changes
- command, README, or extension-index changes

## Verification

- focused store/review/consolidation tests: **133/133 passed**
- long-`TMPDIR` focused run: **133/133 passed**
- post-review store/review tests: **99/99 passed**
- `npm run check`: passed
- `npm run check:min-sdk`: passed against Pi SDK **0.80.1**
- `npm test`: all **42 test files passed**
- `npm pack --dry-run`: passed
- `git diff --check`: clean
- intentional mutation checks confirmed strict-shrink and project-attribution regressions are detected
- independent code review: approved
- final merge gate: approved

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -208,6 +208,8 @@ export async function triggerConsolidation(
           config: llmConfig,
           timeoutMs,
           signal,
+          requireAtomicShrink: true,
+          expectedTarget: toolTarget,
         },
         dbManager,
         projectName,
--- a/src/handlers/review-memory-ops.ts
+++ b/src/handlers/review-memory-ops.ts
@@ -21,6 +21,7 @@ export interface ReviewMemoryOperation {
 export interface ApplyReviewOperationsResult {
   appliedCount: number;
   skippedCount: number;
+  error?: string;
 }
 
 export interface DirectReviewResult {
@@ -36,6 +37,8 @@ export interface RunDirectMemoryCompletionOptions {
   config: Pick<MemoryConfig, "llmModelOverride" | "llmThinkingOverride">;
   timeoutMs?: number;
   signal?: AbortSignal;
+  requireAtomicShrink?: boolean;
+  expectedTarget?: ReviewMemoryOperation["target"];
 }
 
 /** Shared transport gate: review/flush/consolidation/correction all default to
@@ -263,8 +266,64 @@ export async function applyReviewOperations(
   projectStore: MemoryStore | null,
   operations: ReviewMemoryOperation[],
   _dbManager: DatabaseManager | null = null,
-  _projectName?: string | null,
+  projectName?: string | null,
+  options: {
+    requireAtomicShrink?: boolean;
+    expectedTarget?: ReviewMemoryOperation["target"];
+  } = {},
 ): Promise<ApplyReviewOperationsResult> {
+  if (options.requireAtomicShrink) {
+    if (operations.length === 0) {
+      return {
+        appliedCount: 0,
+        skippedCount: 0,
+        error: "Atomic plan requires at least one operation.",
+      };
+    }
+
+    const target = operations[0]?.target;
+    if (!target || operations.some((operation) => operation.target !== target)) {
+      return {
+        appliedCount: 0,
+        skippedCount: operations.length,
+        error: "Atomic plan must use exactly one target.",
+      };
+    }
+    if (options.expectedTarget && target !== options.expectedTarget) {
+      return {
+        appliedCount: 0,
+        skippedCount: operations.length,
+        error: `Atomic plan targeted '${target}', expected '${options.expectedTarget}'.`,
+      };
+    }
+    if (target === "project" && !projectStore) {
+      return {
+        appliedCount: 0,
+        skippedCount: operations.length,
+        error: "Project memory is unavailable.",
+      };
+    }
+
+    const activeStore = target === "project" ? projectStore! : store;
+    const memoryTarget = target === "project" ? "memory" : target;
+    const mutationOperations = operations.map((operation) => ({
+      action: operation.action,
+      content: operation.content,
+      oldText: operation.old_text,
+      category: operation.category,
+      failureReason: operation.failure_reason,
+      project: target === "failure" ? projectName ?? undefined : undefined,
+    }));
+    const result = await activeStore.applyMutationPlan(memoryTarget, mutationOperations, { requireShrink: true });
+    return result.success
+      ? { appliedCount: operations.length, skippedCount: 0 }
+      : {
+          appliedCount: 0,
+          skippedCount: operations.length,
+          error: result.error ?? "Atomic memory plan failed.",
+        };
+  }
+
   let appliedCount = 0;
   let skippedCount = 0;
 
@@ -290,6 +349,7 @@ export async function applyReviewOperations(
           result = await activeStore.addFailure(op.content, {
             category,
             failureReason: op.failure_reason,
+            project: projectName ?? undefined,
           });
           if (result.success) {
             appliedCount++;
@@ -434,14 +494,26 @@ export async function runDirectMemoryCompletion(
       return { ok: true, appliedCount: 0, fallbackReason: "empty" };
     }
 
-    const { appliedCount } = await applyReviewOperations(
+    const applied = await applyReviewOperations(
       store,
       projectStore,
       operations,
       dbManager,
       projectName,
+      {
+        requireAtomicShrink: options.requireAtomicShrink,
+        expectedTarget: options.expectedTarget,
+      },
     );
-    return { ok: true, appliedCount };
+    if (applied.error) {
+      return {
+        ok: false,
+        appliedCount: 0,
+        fallbackReason: "provider_error",
+        error: applied.error,
+      };
+    }
+    return { ok: true, appliedCount: applied.appliedCount };
   } catch (err) {
     if (controller.signal.aborted) {
       return { ok: false, appliedCount: 0, fallbackReason: "aborted" };
--- a/src/handlers/review-memory-ops.ts
+++ b/src/handlers/review-memory-ops.ts
@@ -310,7 +310,7 @@ export async function applyReviewOperations(
       action: operation.action,
       content: operation.content,
       oldText: operation.old_text,
-      category: operation.category,
+      category: target === "failure" ? operation.category ?? "failure" : operation.category,
       failureReason: operation.failure_reason,
       project: target === "failure" ? projectName ?? undefined : undefined,
     }));
--- a/src/store/memory-store.ts
+++ b/src/store/memory-store.ts
@@ -352,7 +352,12 @@ export class MemoryStore {
             : content;
           const scanError = scanContent(normalizedContent);
           if (scanError) return { success: false, error: scanError };
-          if (plannedEntries.some((entry) => this.decodeEntry(entry).text === normalizedContent)) {
+          const normalizedProject = operation.project?.trim() || null;
+          if (plannedEntries.some((entry) => {
+            const decoded = this.decodeEntry(entry);
+            return decoded.text === normalizedContent
+              && (target !== "failure" || decoded.project === normalizedProject);
+          })) {
             return { success: false, error: "Memory mutation plan would add a duplicate entry." };
           }
           plannedEntries.push(this.encodeEntry(normalizedContent, today, today, operation.project));
--- a/src/store/memory-store.ts
+++ b/src/store/memory-store.ts
@@ -25,7 +25,15 @@ import {
   MEMORY_FILE,
   USER_FILE,
 } from "../constants.js";
-import type { MemoryConfig, MemoryResult, MemorySnapshot, ConsolidationResult, MemoryCategory, MemoryOverflowStrategy } from "../types.js";
+import type {
+  MemoryConfig,
+  MemoryResult,
+  MemorySnapshot,
+  ConsolidationResult,
+  MemoryCategory,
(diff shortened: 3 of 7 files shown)


chandra447 (changes requested):
Thank you for this — and I want to be clear up front that the diagnosis in your PR body is right, the auth handling you wrote is careful, and `applyMutationPlan` is a genuinely good idea I'd like to take. This is real engineering effort and I appreciate it.

But I can't merge it, and the reason is the uncomfortable one: **I don't think it fixes #139.**

## The blocker

`createMemoryCompletionRuntime` resolves auth through `registry.getProviderAuth(...)` (`review-memory-ops.ts:113-114`). On 0.83.0 that chains `getProviderAuth` → `ModelRuntime.getAuth` → `Models.getAuth` → `resolveProviderAuth` → `RuntimeCredentials.read` → `AuthStorage.read`, and `AuthStorage.read` is:

```js
const credential = this.data[provider];   // dist/core/auth-storage.js:182-189
```

`this.data` is the snapshot loaded by the single `this.reload()` in the `AuthStorage` constructor (`:147`). Only the OAuth branch (`resolveStoredOAuth`) ever calls `credentials.modify()`, which is the one path that re-reads `auth.json` under lock. The `type: "api_key"` branch from #139 never does.

Verified empirically against **your own SDK version, 0.83.0**: write `auth.json` with `KEY_ONE`, construct `AuthStorage`, rewrite `auth.json` externally with `KEY_TWO_ROTATED`, then `await storage.read("opencode-go")` still returns `KEY_ONE`. Call `storage.reload()` first and it returns `KEY_TWO_ROTATED`.

Resolving *more often* doesn't help when every resolution reads the same cached object. The missing call is `reload()`, and the PR never makes it.

It's slightly worse than a no-op, unfortunately: `review-memory-ops.ts:426` now forwards `apiKey: auth.auth.apiKey` explicitly into the request, and pi-ai's `applyAuth` documents that "explicit request options win per-field" — so a custom rotation provider with its own `streamSimple`, which previously could have resolved its own credential, now receives the stale one.

**The test can't catch this** because it encodes the false premise. `tests/handlers/memory-completion.test.ts:152-158` supplies a `getProviderAuth` mock that deliberately returns a different key per call:

```js
getProviderAuth: async () => { const apiKey = keys[authLookups] ?? keys[keys.length - 1]; authLookups += 1; return { auth: { apiKey } }; }
```

That's exactly the property the real `AuthStorage` does *not* have. The test passes against any implementation that calls `getProviderAuth` twice, so it certifies nothing about the bug. A test that defends this contract has to drive a real `AuthStorage` over a temp `auth.json` rewritten between calls.

## Second blocker: `npm run check` fails on a clean install

`devDependencies` moves `pi-coding-agent`/`pi-ai` to `^0.83.0` (package.json:53-54) but `dependencies` leaves `"@earendil-works/pi-tui": "^0.80.2"` (:61). pi-coding-agent@0.83.0 needs pi-tui `^0.83.0`, so npm installs both — root `0.80.10` plus a nested `0.83.0` — and tsc rejects the structurally distinct classes:

```
src/handlers/skills-command.ts(1295,13): error TS2345: Argument of type 'TUI' is not assignable to parameter of type 'TUI'.
  Types have separate declarations of a private property 'previousLines'.
```

Reproduced under npm 11.12.1 and npm 10.9.9 (the npm bundled with node 22, which CI pins). `main` type-checks clean, so this is introduced here. One-line fix: bump pi-tui to `^0.83.0`.

This never showed up because **CI has never run on your branches** — a first-time-contributor approval gate on my side, not anything you did. I'll clear it so you get signal.

## Unrelated changes I'd want split out regardless

Per repo convention a fix stays scoped to the issue it closes. Beyond auth this PR also:

1. **Removes the direct→subprocess fallback in all four handlers.** `triggerConsolidation` now hard-returns on any non-applying direct result (`auto-consolidate.ts:125-131`) instead of falling through to `pi -p`. Since `reviewTransport` defaults to `direct`, a transient provider error during over-capacity consolidation now *rejects the user's memory write* where it previously recovered. Combined with `requireAtomicShrink: true` (:149) that's a real availability trade-off. Defensible, but it needs its own PR and an explicit statement of the trade.
2. **Deletes the "💾 Memory auto-reviewed and updated" notification** and both `shouldNotifyDirect`/`shouldNotifySubprocess`, leaving `const directResult = await runDirectReview(...)` assigned-and-unread followed by a bare `return`, and `void subprocessResult;` (`background-review.ts:210-224`). This is the path where the extension autonomously writes conversation content to `~/.pi/agent/` — removing the only user-visible signal that it happened, and swallowing the failure reason, is a transparency regression I don't want. The tests were inverted rather than removed, which hid it.
3. **Makes `session_before_compact` propagate flush errors.** `flushSnapshot` dropped the `try { } catch { }` around `runDirect` (`session-flush.ts:76-91`), and `resolveProviderAuth` throws `ModelsError` on credential faults, so a real credential problem now rejects out of the compaction handler. Shutdown kept its `.catch(() => {})`; only compact regressed.
4. **Raises the peer floor `>=0.74.0` → `>=0.81.0`**, forced by `getProvider`/`getProviderAuth` not existing before 0.81. Every user on Pi 0.74–0.80 loses the extension.
5. **Repoints existing tests to `reviewTransport: "subprocess"`** (`auto-consolidate.test.ts:23`, `correction-detector.test.ts:301`, `session-flush.test.ts:112`) rather than adapting them, so the suite passes while covering less of the default configuration than before.

Minor, but worth a line: `applyMutationPlan` honours `operation.project` (`memory-store.ts:359`) but `applyReviewOperations` never populates it (`review-memory-ops.ts:267-271`), so project scope is silently dropped on `add`. Given `areDistinctScopedFailureCopies` exists to disambiguate per-project entries, that scope is load-bearing.

## Where this leaves us

I've opened **#146**, which fixes #139 in two files using `modelRegistry.authStorage.reload()` before `getApiKeyAndHeaders` — both public in the pinned 0.80.2, no dependency change, no peer floor bump — plus a retry on auth rejection when the key actually changed. It type-checks against the pinned SDK.

Proposing: I merge #146 for the bug, and close this in favour of it. **But I'd really like `applyMutationPlan` as its own PR** — atomic all-or-nothing consolidation plans with `scanContent()` on both write branches, inside `runTargetMutation`, is better than what's on main today, and a security pass specifically confirmed it gates correctly and holds the cross-process lock. Same for the explicit-transport change if you want to make the availability argument on its own terms.

Please push back if you think I've got the `AuthStorage.read` chain wrong — I'd rather be corrected than merge #146 on a bad reading.

chandra447:
CI is running now — that was a first-time-contributor approval gate on my side, not anything you did. Sorry for leaving you without signal on three PRs.

**Type Check fails**, exactly as described in my review, verbatim from the runner:

```
src/handlers/skills-command.ts(1295,13): error TS2345: Argument of type
  '.../node_modules/@earendil-works/pi-coding-agent/node_modules/@earendil-works/pi-tui/dist/tui').TUI'
  is not assignable to parameter of type
  '.../node_modules/@earendil-works/pi-tui/dist/tui').TUI'.
  Types have separate declarations of a private property 'previousLines'.
Process completed with exit code 2
```

One line to fix: `@earendil-works/pi-tui` is still `^0.80.2` in `dependencies` while the others moved to `^0.83.0`. Tests and Lint both pass.

---

**One correction to my review, in your favour.** I listed the peer floor bump `>=0.74.0` → `>=0.81.0` as a cost — "every user on Pi 0.74–0.80 loses the extension." I went and checked that claim afterwards, and it was wrong: those users never had it working.

`src/handlers/review-memory-ops.ts` imports `@earendil-works/pi-ai/compat`, and that subpath doesn't exist before 0.80.1:

```
$ node -e "import('@earendil-works/pi-ai/compat')"   # against 0.79.10
ERR_PACKAGE_PATH_NOT_EXPORTED
```

The real floor on `main` today is 0.80.1, bisected by type-checking against each published SDK. `>=0.74.0` was fiction, and it had been for a while. So "you're raising a floor nobody validated" would have been a fair answer to that part of my review, and you'd have been right.

I've opened #149, which corrects the floor and adds a CI job that installs the declared minimum and type-checks against it, so this can't drift again. That also means once #149 lands, a floor bump in your PR becomes a normal reviewable change with CI backing it rather than an unverifiable assertion — which should make the 0.81 requirement easier to argue, if the rest of the design holds up.

The blockers in my review are unaffected: the `AuthStorage.read()` cache means #139 isn't fixed, and the pi-tui mismatch fails the build. But I wanted the record corrected on the part I got wrong.

chandra447:
Heads-up so you're not surprised: #146 has landed, which closed #139 automatically. This PR is still open and I have **not** closed it.

To be explicit about why, since "your PR's issue got closed by someone else's" is a bad experience: the blocker was that this branch's mechanism resolves auth through `AuthStorage.read()`, which returns the same constructor-loaded snapshot, so a rotated `api_key` stayed stale. That's a correctness finding, not a preference — I verified it against your own SDK version. #146 is 2 files calling `authStorage.reload()`.

What I'd still like from this PR, and would review promptly:

**`applyMutationPlan`** (`memory-store.ts:341-410`). Atomic all-or-nothing consolidation plans are better than what's on `main` today, and a dedicated security pass confirmed yours gates `scanContent()` on both write branches and holds the cross-process markdown lock correctly. As a standalone PR that's a clear yes from me. One thing to fix on the way: `operation.project` is honoured at `:359` but never populated by `applyReviewOperations` (`review-memory-ops.ts:267-271`), so project scope is silently dropped on `add`.

**The explicit-transport change**, if you want to make the availability argument on its own terms. I pushed back because removing the direct→subprocess fallback means a transient provider error now rejects the user's memory write where it previously recovered — but that's a trade-off worth debating, not a closed question, and it deserves its own thread rather than riding along with an auth fix.

Also worth knowing: #149 landed, which corrects the peer floor to `>=0.80.1` and adds a CI job that installs the declared minimum and type-checks against it. That was your and #141's floor bumps prompting me to check, and finding the declared floor had been wrong for months. So a floor bump in a future PR of yours is now a normal reviewable change with CI behind it.

Three PRs, two real bugs found (the install spec, and prompting the floor discovery). Thanks for the effort here — I'd rather have this than silence.

xz-dev:
Reworked this PR in place from current `main` (`fe35a63`) and force-pushed `007e204`, following the review guidance to retain `applyMutationPlan` as a standalone change.

The old provider-auth implementation and unrelated transport/dependency/notification/flush changes have been removed. Auth freshness remains the upstream #146 implementation, and the minimum SDK remains `>=0.80.1` from #149.

This branch now contains only the atomic memory mutation primitive, complete same-target consolidation-plan application, strict shrink/expected-target validation, and project-scoped failure-memory propagation. Existing direct→subprocess fallback behavior is preserved.

Verification: focused 133/133, long-TMPDIR 133/133, post-review 99/99, `npm run check`, minimum-SDK 0.80.1 check, all 42 test files, pack dry-run, independent review, and final merge gate all passed.

chandra447 (approved):
Reviewed latest head 007e204075a49b0893393274c870d7d654528985. Atomic mutation plans validate and publish all operations under the existing lock, preserve project-scoped failure metadata, retain fallback behavior, and pass type-check, minimum-SDK, tests, and lint.

PR #143, opened 2026-08-02
fix: poll for consolidation lock on contention; GC dead-pid lock rows

Frank-Opus:
# Consolidation lock contention hard-fails memory writes; hung consolidator blocks all sessions; dead-pid locks never GC'd

**pi-hermes-memory 0.9.2 / pi-coding-agent 0.83.0 / macOS / node 25**

## Summary

When a memory consolidation lock is held (by any session, on any target), **every** memory write that exceeds capacity fails hard with `Consolidation already in progress for target '<t>'`. The consolidation lock acquisition fails immediately on first contention, unlike the store's *mutation* locks which poll with a deadline. A hung-but-alive consolidator therefore blocks memory writes **for all sessions** for the full `staleMs` window (~210 s), and the failure is cascaded into the user-visible memory-write error.

## Deterministic repro

1. Simulate a hung-but-alive consolidator by inserting a lock row with a live pid and a fresh timestamp (key = `sha256(<abs path of MEMORY.md>)`, verified against the running store):
```sql
INSERT INTO locks (lock_key, token, pid, incarnation, acquired_at)
VALUES ('memory:memory:a074fd587d019f90a6a079b64ac7ce323ca396e06b5f8ca740c6027fcfc66784',
        'repro', 69955, NULL, strftime('%s','now')*1000);
```
(`pid 69955` is a live process → liveness check passes → not stealable; fresh `acquired_at` → not past `staleMs` → not stealable. This is exactly the state a hung/slow consolidator leaves behind.)

2. Trigger any capacity-exceeding memory add:
```
pi -p "Use your memory tool to save this exact short text to memory (target: memory, content: HELLO-REPRO-12345). Reply with the tool's result message."
```

3. Actual output:
```
⚠️ Auto-consolidation failed for 'memory': Consolidation already in progress for target 'memory'. Skipping duplicate subprocess.
{"success":false,"error":"Memory at 4991/5000 chars. Adding this entry (17 chars) would exceed the limit. Replace or remove existing entries first. Auto-consolidation attempted but failed: Consolidation already in progress for target 'memory'. Skipping duplicate subprocess."}
```

Observed in production: the same warning + write-failure loop repeatedly during context compaction with two concurrent pi sessions.

## Root cause

- `handlers/auto-consolidate.ts` — `tryAcquireConsolidationLock` returns `null` on the **first** contention; `triggerConsolidation` converts that into an immediate hard failure (`Consolidation already in progress ... Skipping duplicate subprocess.`) with no wait/retry.
- **Contrast:** `store/markdown-mutation-lock.ts` (`acquireMarkdownMutationLock`) polls with `MUTATION_WAIT_MS` deadline for the *mutation* locks — so the codebase already has the correct pattern for transient contention; consolidation just doesn't use it.
- `store/atomic-lock-coordinator.ts` — `staleMs = timeoutMs + 30_000` (default 210 s). A live pid is not stealable until stale; the consolidator child has `timeoutMs` 180 s **plus** `retryWithoutOverrides` (another 180 s) = up to 6 min of lock holding, longer than `staleMs` → the lock can be stolen mid-run (fencing), or (worse) the child hangs and blocks every session for 210 s.
- `store/memory-store.ts` (`addWithConsolidation`, ~line 239) — the auto-consolidation failure is merged into the memory-write error string, so a lock-contention hiccup becomes a visible memory-tool failure mid-task.

## Secondary issue: dead-pid lock rows are never GC'd

A `pi -p "/memory-consolidate"` invocation left `user:user:0fb97a…` in the lock table **after its process (pid 85036) had exited**. Rows are only reaped when another session with the *same storage identity* attempts acquisition; orphaned identities leak rows forever (harmless until that identity returns, then a spurious wait/failure).

## Suggested fix (I have a patch ready — PR to follow)

1. **Poll with deadline** in `tryAcquireConsolidationLock` (mirror `acquireMarkdownMutationLock`): wait up to `min(staleMs, ~30 s)` before giving up, so transient contention becomes a brief delay instead of a hard failure.
2. **Opportunistic dead-pid GC** in `AtomicLockCoordinator`: on each `tryAcquire`, delete rows whose pid is dead (with a small grace to avoid racing a live holder's insert).
3. (Optional) When contention is detected, return a *deferred* success for the memory write path so capacity errors don't cascade into tool failures.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1214,6 +1214,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1231,6 +1234,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1248,6 +1254,9 @@
         "riscv64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1265,6 +1274,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1282,6 +1294,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -38,6 +38,12 @@ type ConsolidationLlmConfig = Pick<MemoryConfig, "llmModelOverride" | "llmThinki
 
 const CONSOLIDATION_LOCK_STALE_GRACE_MS = 30000;
 const CONSOLIDATION_LOCK_ENV = "PI_HERMES_CONSOLIDATION_LOCK_DIR";
+// How long to keep polling for a contended consolidation lock before giving up.
+// Mirrors acquireMarkdownMutationLock: transient contention (another session
+// mid-consolidation) should become a short delay, not an immediate hard
+// failure of the memory write that triggered auto-consolidation.
+const CONSOLIDATION_LOCK_CONTENTION_WAIT_MS = 30000;
+const CONSOLIDATION_LOCK_POLL_MS = 100;
 
 interface ConsolidationLock {
   release: () => Promise<void>;
@@ -67,10 +73,21 @@ async function tryAcquireConsolidationLock(
   const root = consolidationLockRoot();
   await fs.mkdir(root, { recursive: true });
   const coordinator = AtomicLockCoordinator.shared(path.join(root, "locks.sqlite"));
-  const lease = coordinator.tryAcquire(
-    consolidationLockKey(target, toolTarget, storageIdentity),
-    { staleMs: Math.max(timeoutMs, 0) + CONSOLIDATION_LOCK_STALE_GRACE_MS },
-  );
+  const staleMs = Math.max(timeoutMs, 0) + CONSOLIDATION_LOCK_STALE_GRACE_MS;
+  const key = consolidationLockKey(target, toolTarget, storageIdentity);
+
+  // Contention is normally transient (another session is mid-consolidation or
+  // a hung holder is about to cross staleMs). Poll with a bounded deadline
+  // instead of failing the caller on the first collision; a hard failure here
+  // cascades into every memory write that exceeds capacity (memory-store.ts
+  // addWithConsolidation), blocking writes for all sessions while the lock is
+  // held.
+  const deadline = Date.now() + Math.min(CONSOLIDATION_LOCK_CONTENTION_WAIT_MS, staleMs);
+  let lease = coordinator.tryAcquire(key, { staleMs });
+  while (!lease && Date.now() < deadline) {
+    await new Promise((resolve) => setTimeout(resolve, CONSOLIDATION_LOCK_POLL_MS));
+    lease = coordinator.tryAcquire(key, { staleMs });
+  }
   return lease ? { release: async () => lease.release() } : null;
 }
 
--- a/src/store/atomic-lock-coordinator.ts
+++ b/src/store/atomic-lock-coordinator.ts
@@ -96,11 +96,17 @@ const RELEASE_ATTEMPTS = 3;
 const pendingReleases = new Map<string, () => void>();
 const sharedCoordinators = new Map<string, AtomicLockCoordinator>();
 
+// Opportunistic GC tuning: how old (ms) a row must be before a dead-pid row
+// qualifies for deletion, and how often each coordinator may sweep.
+const DEAD_LOCK_SWEEP_GRACE_MS = 30000;
+const DEAD_LOCK_SWEEP_INTERVAL_MS = 60000;
+
 export class AtomicLockCoordinator {
   private readonly pid: number;
   private readonly incarnation: string | null;
   private readonly probeIncarnation: (pid: number) => string | null;
   private cachedDb: DatabaseLike | null = null;
+  private lastSweepMs = 0;
 
   constructor(private readonly dbPath: string, options: AtomicLockCoordinatorOptions = {}) {
     this.pid = options.pid ?? process.pid;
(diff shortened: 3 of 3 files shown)


chandra447 (commented):
Reviewed alongside your report in #144 — thank you again for both. The diagnosis is right and the two mechanisms you chose (poll with a deadline, opportunistic dead-pid GC) are the two I ended up with.

I've opened **#145**, which supersedes this. Not a judgement on the patch — three things pushed me toward one branch rather than iterating here:

**1. `staleMs` is wrong in the other direction too, and polling alone doesn't reach it.** Your PR keeps `staleMs = timeoutMs + 30_000` (210s) while the holder can legitimately run for `2 × timeoutMs` (360s) once `retryWithoutOverrides` fires. So a *healthy* consolidator still gets robbed mid-run at 210s and two children write the same store concurrently. #145 heartbeats the lease instead (`AtomicLockCoordinator.renew()`, token-fenced `UPDATE ... acquired_at`, beat every 10s), which decouples `staleMs` from `timeoutMs` entirely and lets it drop to a fixed 45s — so a wedged holder is reclaimed 4.5× faster *and* a slow-but-healthy one can't lose its lease at all.

**2. A 30s poll window is a long time to block a tool call.** `CONSOLIDATION_LOCK_CONTENTION_WAIT_MS = 30000` means the model sits inside a `memory` tool call for up to half a minute on contention. #145 uses 5s (matching `MUTATION_WAIT_MS`) and makes the give-up path informative rather than punishing — `deferred: true` with *"retry in a moment"* instead of a hard "Skipping duplicate subprocess" failure.

**3. Two details in the GC worth carrying forward if you review #145:**

- `DELETE FROM locks WHERE lock_key = ? AND pid = ?` is fenced on pid, which doesn't cover pid reuse: you confirm pid *P* is dead, a new process recycles *P* and acquires the lock, and the sweep deletes its fresh row. #145 fences on `(lock_key, token, acquired_at)` and additionally filters `acquired_at <= now - grace` in SQL, so a row re-taken between the read and the delete is left alone.
- The sweep uses the module-level `processIsAlive` only, bypassing the injectable `probeIncarnation` from `AtomicLockCoordinatorOptions`. Test doubles pass synthetic pids (`4242`) that aren't alive locally — today the 30s grace happens to protect them, but that's luck rather than design. #145 checks `processIsAlive` first (cheap, and true for live pids so no `ps` spawn) then falls back to `this.probeIncarnation`, matching the same liveness predicate `tryAcquire` steals on.

Also: `package-lock.json` picked up `libc` field churn from a different npm version. The `Lint` CI job runs `git diff --exit-code`, so that would have failed — worth a `git checkout package-lock.json` habit before pushing.

Closing this in favour of #145, with the polling and GC shape credited to you in the commit message and in #144. If you have the two-session reproduction environment still set up, a review on #145 from you specifically would be worth more than anyone else's — you're the only person who's actually hit this in production.

chandra447:
CI has run now (that was a first-time-contributor approval gate on my side, sorry). Two failures, and one of them is information my review didn't have:

**Lint — fails.** The `libc` lockfile churn I mentioned. The job is `git diff --exit-code` after `npm install`, and it rejects:

```
-      "libc": [
-        "musl"
-      ],
```

**Tests — fails, and this is the interesting one.**

```
not ok 4 - skips a duplicate subprocess while the same target is consolidating
  failureType: 'cancelledByParent'
  error: 'Promise resolution is still pending but the event loop has already resolved'
```

That single hang cancels the rest of the file — 30 subsequent tests in `auto-consolidate.test.ts` go down with it.

It's a genuine consequence of the change, not flakiness, and it's worth understanding because I hit exactly the same wall. The existing test holds the lock in a fake `pi.exec` that never resolves, races the second `triggerConsolidation` against a 100ms timer, then releases. With instant-fail contention the second call returns before that release. With polling it doesn't — it waits, the release fires, it *acquires* the lock and spawns a second child whose resolver was never captured by the test's `releaseExecs` array. That promise never settles, the event loop drains, node cancels the suite.

The test was encoding "contention means skip" as a premise. Once contention means "wait", it deadlocks. I ran into the identical hang on my first pass at #145 — mine sat for 300s before I killed it.

The fix isn't to the source, it's that the test has to be rewritten to express the new contract. In #145 I split it into two, with the poll window injectable via `PI_HERMES_CONSOLIDATION_LOCK_WAIT_MS` so neither one costs real wall-clock time:

- `defers a duplicate subprocess while the same target is consolidating` — wait window `0`, asserts `deferred: true` and exactly one child spawned
- `waits out transient contention instead of failing the caller` — wait window `2000`, releases the first holder, asserts the second one actually consolidates rather than hard-failing

Flagging it here rather than leaving you to debug a `cancelledByParent` cascade, which is a miserable error to chase — it names the wrong test and hides the real cause.

Everything else in my earlier review stands. Type Check passes.

chandra447:
#145 has landed on `main` (5ce3249), so I'm closing this — but I want to be clear about what that means, because "closed" is the wrong signal for what you actually contributed here.

You found the bug, wrote a deterministic repro with the exact lock key, identified both root causes correctly, spotted that `acquireMarkdownMutationLock` already had the right pattern, and shipped a patch. #145 is your diagnosis plus your two mechanisms — bounded polling and dead-pid GC — with a heartbeat added on top and the contention path reworded. The commit message credits you, and so does #144.

What shipped, for the record:

- **Bounded polling** on contention — your suggestion 1, at 5s rather than 30s so a memory tool call doesn't stall.
- **Dead-pid GC** — your suggestion 2, fenced on `(token, acquired_at)` rather than `pid` to close the pid-reuse window, and using the injectable `probeIncarnation` so test doubles behave.
- **Lease heartbeat** — the piece your patch didn't have. `staleMs` was *also* wrong in the other direction (210s stale vs a 360s worst-case hold), so a healthy consolidator could be robbed mid-run. Beating every 10s lets `staleMs` drop to a fixed 45s: wedged holders reclaim 4.5× faster, and slow-but-healthy ones can't lose the lease at all.
- **`deferred` instead of failure** — your optional suggestion 3, minus the "deferred success" framing, since the write genuinely didn't happen.
- **Skip-redundant-child** — after waiting out contention, re-read from disk and skip our own LLM pass if the session ahead already freed space.

Two things your PR surfaced that I'd have missed otherwise: the `libc` lockfile churn (the `Lint` job is `git diff --exit-code`, worth knowing before your next PR), and the test deadlock CI exposed — which I hit identically on my own first attempt and which forced the rewrite of that test into two.

Thank you for this. A report of this quality is rare, and a report that arrives with a working patch is rarer. If you hit anything else in your two-session setup, please open it — you're clearly running this harder than most.

PR #152, opened 2026-08-04
fix: allowlist suffixless auth adapters in child subprocesses

knthmn:
## Summary

Child `pi -p` subprocesses use `--no-extensions`, then re-add provider auth adapters so
subscription billing headers remain active. Detection only recognized the
`*-auth-adapter` / `*-oauth-adapter` convention, so the installed `pi-claude-auth` adapter was
omitted and Anthropic rejected the child request as third-party extra usage.

## Changes

- Keep the existing naming convention as the general trust boundary.
- Add exact allowlist entries for `pi-claude-auth` and `@gotgenes/pi-anthropic-auth`.
- Pass full `@scope/name` values into the matcher so scoped allowlist entries are exact.
- Test both allowlisted packages and reject an arbitrary `pi-vault-auth` package.
- Derive expected child adapter args in tests so contributors with an installed adapter do not
  get environment-dependent failures.

This closes the reported adapter gap without making a generic `*-auth` suffix sufficient to
load an extension into maintenance children. Future suffixless adapters can use
`childExtensionPaths` until reviewed and allowlisted.

## Verification

- `npm run check`
- `npm run check:min-sdk`
- `npm test` — all 42 test files pass (38 tests in `pi-child-process.test.ts`)
- End-to-end reproduction: the child 400s without `pi-claude-auth`; detection now adds its
  extension path and the same child request returns `OK`

### DIFF
--- a/src/handlers/pi-child-process.ts
+++ b/src/handlers/pi-child-process.ts
@@ -116,8 +116,19 @@ const AUTH_ADAPTER_NAME_PATTERNS: readonly RegExp[] = [
   /(^|[-/])auth-adapter$/,
 ];
 
+// These provider adapters predate the *-auth-adapter convention. Keep this
+// list exact so an arbitrary suffixless *-auth package cannot access child
+// maintenance prompts.
+const AUTH_ADAPTER_PACKAGE_NAMES: ReadonlySet<string> = new Set([
+  "pi-claude-auth",
+  "@gotgenes/pi-anthropic-auth",
+]);
+
 function isAuthAdapterPackageName(name: string): boolean {
-  return AUTH_ADAPTER_NAME_PATTERNS.some((pattern) => pattern.test(name));
+  return (
+    AUTH_ADAPTER_PACKAGE_NAMES.has(name)
+    || AUTH_ADAPTER_NAME_PATTERNS.some((pattern) => pattern.test(name))
+  );
 }
 
 // Read a sibling package's "pi": { "extensions": [...] } manifest field —
@@ -166,7 +177,7 @@ function scanRootForAuthAdapters(root: string): string[] {
         continue;
       }
       for (const scopedName of scopedPackages) {
-        if (!isAuthAdapterPackageName(scopedName)) continue;
+        if (!isAuthAdapterPackageName(`${entry}/${scopedName}`)) continue;
         detected.push(...readPackageExtensionEntries(join(scopeDir, scopedName)));
       }
       continue;
--- a/tests/handlers/pi-child-process.test.ts
+++ b/tests/handlers/pi-child-process.test.ts
@@ -27,7 +27,12 @@ const OWN_EXTENSION_PATH = path.resolve(
   "../../src/index.ts",
 );
 
-const EXT_ARGS = ["--no-extensions", "-e", OWN_EXTENSION_PATH];
+// Adapter detection scans the real sibling/agent node_modules trees, so a
+// contributor with a provider auth adapter installed gets extra -e args in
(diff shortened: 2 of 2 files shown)


chandra447 (changes requested):
Thanks for isolating the residual adapter gap and for documenting the child-prompt trust boundary explicitly. The reproduction and the `pi-claude-auth` coverage are useful, and the branch passes the full local validation set.

One blocker remains before I can approve this:

**The generic suffixless matcher broadens the child extension trust boundary too far.**

`/(^|[-/])o?auth$/` matches every sibling package directory ending in a delimited `auth` or `oauth`. `scanRootForAuthAdapters()` searches both the extension sibling root and `~/.pi/agent/npm/node_modules`, including every scope. `readPackageExtensionEntries()` only requires a valid `pi.extensions` entry; it does not verify that the package is enabled in the parent. Every match is then passed beside `--no-extensions` as `-e`, so an installed-but-disabled package such as `pi-vault-auth` would newly execute in review/flush/correction/consolidation children and receive the private `@.../prompt.md` payload.

I reproduced that exact shape: a fixture named `pi-vault-auth` with `{ "pi": { "extensions": ["index.ts"] } }` is returned by `detectAuthAdapterExtensionPaths()` on this branch. That is broader than the existing deliberate `*-auth-adapter` convention and conflicts with the isolation boundary that excludes arbitrary inherited extensions.

Smallest safe fix:

- keep the existing `*-auth-adapter` / `*-oauth-adapter` convention;
- add exact full-package allowlist entries for the two verified suffixless adapters: `pi-claude-auth` and `@gotgenes/pi-anthropic-auth`;
- pass the full `@scope/name` into the matcher for scoped packages;
- add positive tests for both real package names and a negative `pi-vault-auth`-style suffixless package.

Users can still use `childExtensionPaths` for future suffixless adapters until they are reviewed and added. That closes the reported gap without making package-name suffix alone sufficient to access maintenance prompts.

Validation on head `232fde2`: `npm run check`, `npm run check:min-sdk`, and all 42 test files pass. The blocker is the newly expanded execution boundary, not CI correctness.

knthmn:
Addressed in the new commit

- Restored the existing `*-auth-adapter` / `*-oauth-adapter` convention.
- Added exact allowlist entries for `pi-claude-auth` and `@gotgenes/pi-anthropic-auth`.
- Passed full `@scope/name` values into the matcher.
- Added positive tests for both allowlisted packages and a negative `pi-vault-auth` test.

All vefirications are passing. The PR description has been updated to reflect the narrower implementation.

chandra447:
Focused review of current head b1cf5fd confirms the prior trust-boundary blocker is addressed: convention matches remain, the two suffixless adapters are exact-allowlisted, scoped names are matched fully, pi-vault-auth-style packages are rejected, and the focused child-process suite passes 38/38. Please rerun CI and re-review the updated head.

chandra447:
Final follow-up on b1cf5fd: the requested trust-boundary fix is present. The implementation preserves only the *-auth-adapter/*-oauth-adapter conventions, exact-allowlists pi-claude-auth and @gotgenes/pi-anthropic-auth, matches full scoped names, and rejects pi-vault-auth-style suffixless packages. Local verification passes: npm run check and the focused child-process suite 38/38. Please re-review and merge when CI is green.

chandra447:
Thank you for raising this adapter compatibility fix and for tightening the trust boundary after review. The exact allowlist, scoped-name handling, and pi-vault-auth regression coverage address the blocker; CI is green, so I am proceeding with the merge.

PR #154, opened 2026-08-05
feat(session_search): custom TUI renderers for call/result

4cya:
Adds renderCall/renderResult for both anchor and legacy modes so the session_search tool output is readable in the TUI:

- call header: compact one-liner showing the query and active filters
- collapsed result: count + truncation warnings (N snippets truncated, output capped) instead of dumping the full text block
- expanded result: first 25-30 lines with '(N more lines)' tail
- isPartial: 'Searching...' spinner
- error/no-results: dim message

Pure display-layer change. The model still receives the same result.content[0].text. The same renderers drive /export HTML output.

### DIFF
--- a/src/tools/session-search-tool.ts
+++ b/src/tools/session-search-tool.ts
@@ -8,6 +8,7 @@ import { searchSessionAnchors } from '../store/session-anchor-search.js';
 import type { SessionAnchorRange, SessionAnchorSearchResult } from '../store/session-anchor-search.js';
 import type { SessionSearchConfig } from '../types.js';
 import { AGENT_ROOT } from '../paths.js';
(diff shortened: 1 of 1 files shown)


chandra447:
Thank you for contributing the session_search renderer work and for focusing on the TUI experience. I am not merging this PR independently because PR #141 already landed the shared renderer path, while this branch duplicates that work and bypasses the shared display sanitizer. Please close/supersede this branch, or send a narrow follow-up on current main that reuses the shared renderer and sanitizer.

4cya:
Closing per maintainer feedback — superseded by #141's shared output view. Thanks for the review.

Issue #135, opened 2026-07-28
Auto-consolidation failures are silently swallowed — addWithConsolidation discards consolidation.error


jasonrr:
Follow-up to #105, whose "Expected behavior" list included *"Surface an explicit failure to the parent agent rather than silently continuing"* and *"Make auto-spawned consolidation work observable."* In v0.9.1 that part appears unimplemented.

**Observed:** A `memory` tool write hit the 5000-char limit. The tool returned `Memory at 4808/5000 chars...` ~60s after the call. The auto-consolidator had fired and failed — but no failure reason surfaced anywhere: not in the tool result, not in logs, not in the session.

**Cause:** `triggerConsolidation` in `src/handlers/auto-consolidate.ts` returns a detailed error string on every failure path (lock contention: `"Consolidation already in progress..."`, subprocess non-zero exit with stderr, timeout termination via `describeConsolidationFailure`). But `src/store/memory-store.ts:244-253` discards it:

```ts
try {
  const consolidation = await this.consolidator(target, signal);
  if (consolidation.consolidated) { /* reload + retry */ }
} catch {
}
return result;  // original "Memory at ..." error, unchanged
```

Every failure mode (lock contention, spawn failure, non-zero exit, timeout kill) presents identically as the original capacity error, making the auto path impossible to diagnose from the outside.

**Suggested fix:** When `consolidation.consolidated` is false, append `consolidation.error` to the returned error (e.g. `"... Replace or remove existing entries first. Auto-consolidation attempted but failed: <reason>"`), and/or `console.error` it. In our case the swallowed reason was a 60s timeout kill — one line of logging would have saved a source-diving session.

chandra447:
Thank you for this one, @jasonrr — and thank you for using the extension. Fixed in [#138](https://github.com/chandra447/pi-hermes-memory/pull/138), shipping in 0.9.2.

You diagnosed this better than the code did. Pointing straight at `memory-store.ts:244-253` and quoting the `catch {}` left nothing to guess at, and the observation that *every* failure mode — lock contention, spawn failure, non-zero exit, timeout kill — presents identically to a plain full memory was the actual bug. The reason string already existed; we just dropped it on the floor.

What changed:

- The reason is appended to the tool result, exactly as you suggested: `... Replace or remove existing entries first. Auto-consolidation attempted but failed: <reason>`
- It is also logged once where the consolidator is wired, so it reaches the session output and not only the model's view of the tool result.
[...]

PR #165, opened 2026-08-08
feat(memory): configure auto-consolidation failure warnings

xz-dev:
## Summary

- add `autoConsolidationWarnOnFailure`, defaulting to `true`
- allow users to suppress only the session-console warning emitted after a failed automatic consolidation
- preserve the failed tool result, automatic consolidation behavior, manual consolidation reporting, timeout startup warnings, SQLite synchronization warnings, and all other errors

## Context

Issue #135 and PR #138 intentionally made failed automatic consolidation observable both in the tool result and through a session-console warning. This keeps that default while adding a narrow opt-out for users who still want mandatory tool-result diagnostics but do not want the repeated console warning.

## Validation

- `npm run check`
- `npx tsx --test tests/config.test.ts tests/index.test.ts` (33 passed)
- mutation check: forcing the warning predicate to always return false fails the default-warning test
- independent review: approved with no functional, minimality, test, or documentation findings

The full suite could not be attested locally because this machine runs Node 26, while `better-sqlite3@12.9.0` supports Node 20–25 and has no Node ABI 147 binding. The changed-area tests and TypeScript check are green; CI can validate on the repository-supported runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,6 +508,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
   "failureInjectionMaxAgeDays": 7,
   "failureInjectionMaxEntries": 5,
   "consolidationTimeoutMs": 180000,
+  "autoConsolidationWarnOnFailure": true,
   "flushOnCompact": true,
   "flushOnShutdown": true,
   "flushMinTurns": 6,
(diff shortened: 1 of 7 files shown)


chandra447:
Review verdict: approve/merge. Current CI is green (Type Check, Min Supported SDK, Tests, Lint). The option is backward-compatible, defaults to the existing warning behavior, and gates only the automatic-consolidation session-console warning; failed tool results, manual consolidation output, startup warnings, and SQLite warnings remain observable. This is a good narrow Pi extension configuration change. It is not a fix for the silent background-review provider failure in issue #169.

PR #171, opened 2026-08-12
fix(auto-consolidation): defer stale extension ctx errors instead of reporting failure

21307369:
## Problem

When a memory write triggers auto-consolidation and the session is replaced or reloaded while the consolidation child process is still running (`newSession`/`fork`/`switchSession`/`reload`), the captured extension `pi` object is invalidated by the framework. The in-flight `pi.exec` call then throws:

```
This extension ctx is stale after session replacement or reload. Do not use a captured pi or command ctx after ctx.newSession(), ctx.fork(), ctx.switchSession(), or ctx.reload().
```

`triggerConsolidation` caught this and reported it as a **failure**:
- `⚠️ Auto-consolidation failed for 'failure': Consolidation failed: Error: This extension ctx is stale...` (session console warning, `index.ts`)
- the triggering memory write returned an "Auto-consolidation attempted but failed" error

This is noise, not data loss: the new session re-initializes the memory store and consolidates on its own next write.

## Fix

Detect the stale-ctx error message in `triggerConsolidation`'s catch block and return a **deferred** result instead of a failure. The caller (`addWithConsolidation` / `runAutoConsolidation`) already treats deferred as "retry shortly" — no failure warning, no failed tool result.

## Changes

- `src/handlers/auto-consolidate.ts`: recognize `extension ctx is stale` errors → `{ consolidated: false, deferred: true, error: "session replaced or reloaded during consolidation — will consolidate on next write" }`
- `tests/handlers/auto-consolidate.test.ts`: new test verifying a stale-ctx `pi.exec` throw defers instead of failing

## Validation

- `npx tsc --noEmit` clean
- `tests/handlers/auto-consolidate.test.ts`: 35/35 pass (incl. new test)
- `tests/store/memory-store.test.ts`: 78/78 pass (deferred path unaffected)

### DIFF
--- a/src/handlers/auto-consolidate.ts
+++ b/src/handlers/auto-consolidate.ts
@@ -274,10 +274,22 @@ export async function triggerConsolidation(
       consolidated: false,
       error: describeConsolidationFailure(result, timeoutMs),
     };
-  } catch (err) {
+} catch (err) {
+    const message = String(err);
+    if (message.includes("extension ctx is stale")) {
+      // Session replaced/reloaded while consolidation was running. The new
+      // session re-initializes the store and will consolidate on its own next
+      // write, so this is a skip, not a failure — report it as deferred so the
+      // caller asks for a retry instead of surfacing a stale-ctx error.
+      return {
+        consolidated: false,
+        deferred: true,
+        error: "session replaced or reloaded during consolidation — will consolidate on next write",
+      };
+    }
     return {
       consolidated: false,
-      error: `Consolidation failed: ${String(err).slice(0, 200)}`,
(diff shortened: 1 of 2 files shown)


chandra447:
Review verdict: approve/merge after the focused review. Current CI is green (Type Check, Min Supported SDK, Tests, Lint), and classifying Pi stale-context failures as deferred matches the installed SDK lifecycle contract: captured contexts become invalid after session replacement/reload. Please keep the match narrowly tied to the framework error and retain the generic failure path for unrelated errors. Add or retain coverage for both Error objects and non-Error throws if the handler can receive them. This removes misleading failure output but does not solve the separate overflow/consolidation race in issue #177.

Issue #174, opened 2026-08-13
FTS5 search silently misses pure-CJK queries (unicode61 doesn't segment CJK)


Xctickle:
## Summary

`memory_search` (and session search) silently returns **no results for pure-CJK queries**, because the FTS5 index uses the default `unicode61` tokenizer, which does not segment Chinese/Japanese/Korean text. This makes a large class of memories effectively unsearchable for CJK users.

## Reproduction

Stored a memory mixing CJK and English, e.g. `设备清单(devices) … NAS … cdm-xxc …`.

| query | match |
|---|---|
| `设备` (2-char CJK) | ❌ none |
| `设备清单` (4-char CJK) | ❌ none |
| `NAS` / `cdm-xxc` (English) | ✅ hits |

Pure-CJK substrings never match; only embedded English tokens do.

## Root cause

`src/store/schema.ts` creates the FTS5 tables without specifying a tokenizer:

```sql
CREATE VIRTUAL TABLE memory_fts USING fts5(content, content='memories', content_rowid='id');
CREATE VIRTUAL TABLE message_fts USING fts5(content, content='messages', content_rowid='rowid');
```

The default `unicode61` tokenizer treats consecutive CJK characters as one opaque token, so CJK substrings can't be matched by FTS.

## Environment

- pi-hermes-memory `0.9.4`
- better-sqlite3 `12.11.1` (SQLite 3.45 — has built-in `trigram`)
- Node 24

## Suggested fix

**Option A — built-in trigram tokenizer** (simplest):
```sql
… USING fts5(content, content='memories', content_rowid='id', tokenize='trigram')
```
Trigram handles CJK substrings well (≥3 chars). Verified locally: after switching, `设备清单` matches. Minor limitation: 1–2 char queries still miss.

**Option B — make the tokenizer configurable**, defaulting to `trigram` (or at least document the CJK tradeoff).

Either way, existing DBs need a one-time migration (`DROP` + recreate + repopulate), since `CREATE … IF NOT EXISTS` won't rebuild existing tables.

## Why it's easy to miss

The failure is silent (empty results, no error), and most technical memories contain enough English (code, paths, tool names, IPs) that English-token matches mask the CJK gap — until a user searches a pure-CJK term.

Happy to send a PR if this direction sounds right.

chandra447:
[...]

PR #175, opened 2026-08-13
feat: optional qmd semantic search with FTS5 fallback

isaacnewtonagent:
## What does this PR do?

Adds opt-in **qmd** (`@tobilu/qmd`) semantic search to `memory_search`. qmd is a local, on-device search engine for markdown (BM25 keyword + vector semantic + LLM rerank via node-llama-cpp GGUF models) — no cloud, no data leaves the machine.

When `qmdSearch: true` is set in the extension config and the `qmd` CLI is on PATH:
- `memory_search` tries qmd first (meaning-based recall that FTS5 keyword matching misses)
- Falls back to the existing SQLite FTS5 path on any failure (qmd missing, binary error, timeout)
- Memory writes trigger a fire-and-forget qmd reindex via the `tool_result` event

Default is **off** — existing behavior is unchanged for all current users.

## How did you verify your code works?

- `bun run check` (tsc --noEmit): clean
- `bun run test`: all 46 test files pass, including a new test that qmd-enabled search falls back to FTS5 when the qmd binary is unavailable
- The qmd client is fully best-effort: detection is cached, every call is timeout-bounded, and failures never surface to the tool result

## Checklist

- [x] Tests pass
- [x] Typecheck passes
- [x] No breaking changes (feature is opt-in, default off)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -63,6 +63,7 @@ const DEFAULT_CONFIG: MemoryConfig = {
   standingInstructionsEnabled: true,
   projectsMemoryDir: DEFAULT_PROJECTS_MEMORY_DIR,
   sessionSearch: { variant: "legacy" },
+  qmdSearch: false,
 };
 
 export const DEFAULT_CONFIG_PATH = path.join(
@@ -130,6 +131,7 @@ export function loadConfig(configPath = DEFAULT_CONFIG_PATH): MemoryConfig {
       if (typeof parsed.failureInjectionMaxEntries === "number") config.failureInjectionMaxEntries = parsed.failureInjectionMaxEntries;
       if (typeof parsed.nudgeToolCalls === "number") config.nudgeToolCalls = parsed.nudgeToolCalls;
       if (typeof parsed.standingInstructionsEnabled === "boolean") config.standingInstructionsEnabled = parsed.standingInstructionsEnabled;
+      if (typeof parsed.qmdSearch === "boolean") config.qmdSearch = parsed.qmdSearch;
       if (typeof parsed.projectCharLimit === "number") config.projectCharLimit = parsed.projectCharLimit;
       if (typeof parsed.memoryDir === "string") {
         const normalizedMemoryDir = normalizeConfiguredMemoryDir(parsed.memoryDir);
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,6 +57,7 @@ import { buildPromptContext } from "./prompt-context.js";
 import { migrateLegacyProjectMemoryDirs } from "./project-memory-migration.js";
 import { AGENT_ROOT } from "./paths.js";
 import { isDatabaseMigrationPending } from "./extension-root-migration.js";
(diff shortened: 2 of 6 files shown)


chandra447:
Review verdict: hold/request changes; CI is green, but the current integration is not yet equivalent to memory_search and is not a complete fix for issue #174. Specific gaps: qmd is queried against config.memoryDir only, while the tool supports project/user/failure/category scope; returned paths/snippets are not mapped back to the actual memory records; deletes/replaces can leave stale index hits; and reindex calls can overlap after every memory_* tool result. The qmd CLI invocation must be verified against a pinned/documented runtime prerequisite, including embed/update semantics and JSON shape. For Pi compatibility, prefer the supported pi.exec surface with explicit cwd, timeout, and abort handling instead of an unmanaged child_process call, or explain why the direct call is required. Add success-path mocks for CLI args/output, scope filtering, malformed JSON, timeout, failed reindex, and concurrent writes. Keep the default FTS5 path unchanged and treat qmd as an optional enhancement, not the CJK correctness fix.

chandra447:
Thanks for contributing opt-in qmd semantic search and the FTS5 fallback. The current branch is dirty and conflicts with `main`, so it is not merge-ready. Rebase onto current `main`, resolve the conflict, and rerun the typecheck, tests, and fallback coverage on the new head. Once the branch is current, I will review the qmd timeout, indexing lifecycle, and default-off behavior again.

chandra447 (commented):
Adversarial review (4 independent reviewers, all converged) — 4 blocking issues, two of them independently fatal to the feature:

1. **Double registration kills the qmd path** (src/index.ts). The new §4b `registerMemorySearchTool(pi, dbManager, config)` is overwritten by the unchanged §11 `registerMemorySearchTool(pi, dbManager)`. Last registration wins, so the qmd branch never executes in production (the unit test only exercises the first registration). Fix: single registration, pass config through.
2. **`config.memoryDir` is unset by default** (src/index.ts:245,251). `memoryDir` is only set when the user configures it; the store lives in the computed `globalDir`. With the default layout `qmdDir`/`dir` is `undefined`, so search always falls through to FTS5 and the reindex hook never runs even with `qmdSearch: true`. Fix: thread the resolved global dir, not `config.memoryDir`.
3. **qmd hit path ignores tool filters** (src/tools/memory-search-tool.ts:70-83). On qmd success it returns file paths/snippets immediately, dropping `project`/`target`/`category` — a filtered search can return unscoped results, and the structured entries `memory_replace`/`memory_remove` need are skipped.
4. **Reindex hook fires on searches** (src/index.ts:241-252). `toolName.startsWith('memory_')` includes `memory_search` (success:true even on zero hits), so every search spawns a 30s fire-and-forget `qmd embed` with no debounce/abort — overlapping embeds under chatty agents. Gate on `memory_add`/`replace`/`remove`, debounce/coalesce, and pass the tool abort signal into `execFile` (currently timeout-only, so aborted calls leak 15-30s child processes).

Issue #177, opened 2026-08-13
Auto-consolidation is not coordinated with the active agent — a context-less background pass can rewrite in-flight edits


dimalo:
### Summary

When a memory target reaches its character limit, the active agent's write is rejected, and auto-consolidation can be triggered independently on the same target. The two paths are not coordinated: the background consolidator runs with a snapshot of context that may not reflect the agent's in-flight edits, and it can act on the same target the agent is still editing.

The result is that a background pass — holding less context than the main agent — can rewrite or remove entries in a way the main agent would not have chosen.

### Expected

On overflow, the main agent should get a clear signal and a grace window to consolidate manually (with full context) before any background consolidation runs. The background consolidator should be suppressed during that window.

### Actual

The agent receives a generic limit error with no consolidation-status signal, and auto-consolidation can fire at any time, racing the agent's manual edits.

### Proposed fix

Suppress auto-consolidation for a grace window after an overflow, so the main agent (with full context) gets to consolidate manually first:

- on overflow, mark the target and reject with a "consolidate first" message;
- suppress auto-consolidation while the grace period is active;
- clear the overflow state on the next successful write.

Open design question: wall-clock vs turn-based grace period. The branch implements wall-clock (`overflowGraceMs`, default 180000 ms), because turn counts are only tracked in the background-review handler and aren't reliably observable from the store's overflow path — but this is worth discussing.

Implemented on `dimalo/pi-hermes-memory:feat/overflow-debounce-and-routing` (no PR).

### Related

- #144 (closed) — consolidation lock contention
- #135 / #136 (closed) — auto-consolidation failure handling / timeout

---

_Disclosure: this issue was discussed with and edited by a Pi coding agent (AI)._

chandra447:
Review update: confirmed this is a real data-integrity race. The overflow path can invoke auto-consolidation immediately after returning the limit error, while the active agent may still be preparing a manual replace/add. The store lock only serializes writes; it does not coordinate intent. Recommended fix: add per-target overflow state with an explicit grace policy, suppress background consolidation during that window, and clear the state after a successful write. Please add tests for overflow -> manual edit and overflow -> background attempt, including concurrent/retry behavior. PR #178 improves same-turn curation by returning entries, but does not solve this coordination race.

chandra447:
[...]

PR #178, opened 2026-08-14
feat(memory-store): include current entries in memoryFullError response

jingyulong:
## Problem

When memory is at capacity and `memoryOverflowStrategy` is `"reject"` (or when `"fifo-evict"` cannot fit the new entry alone), `memoryFullError` returns only a plain error string:

> Memory at 4991/5000 chars. Adding this entry (235 chars) would exceed the limit. Replace or remove existing entries first.

The agent then has to issue a separate read of the memory file to see the current entries before it can curate (replace/remove then retry). This is the exact friction the [Hermes Agent memory design](https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/features/memory.md) avoids by including `current_entries` in the full error response.

## Solution

Populate the existing optional `entries`, `usage`, and `entry_count` fields on `MemoryResult` in the error path — they were already defined in the type and used by `successResponse`, just not filled by `memoryFullError`.

The `entries` array contains decoded entry text (metadata stripped via `decodeEntry`), so the agent gets a clean list it can act on directly.

The error message is also updated to guide the agent: *"see the entries list below, then retry this add — all in this turn"* — matching the curated-memory path where the agent consolidates and retries in the same turn.

## Why no tool-layer change?

`formatMemoryToolText` in `memory-tool.ts` falls through to `JSON.stringify(result)` when `success === false`, so the new fields are transparently serialized to the LLM with zero changes outside `memory-store.ts`.

## Verification

Tested with a real at-capacity memory store (4991/5000 chars, 7 entries). The `memory_add` call now returns:

```json
{
  "success": false,
  "error": "Memory at 4991/5000 chars. Adding this entry (235 chars) would exceed the limit. Replace or remove existing entries first (see the entries list below), then retry this add — all in this turn.",
  "target": "memory",
  "usage": "4991/5000 chars",
  "entry_count": 7,
  "entries": ["<entry 1 text>", "<entry 2 text>", "..."]
}
```

The agent was able to `memory_replace` an overlapping entry and retry the `add` in the same turn, without a separate file read.

### DIFF
--- a/src/store/memory-store.ts
+++ b/src/store/memory-store.ts
@@ -348,9 +348,14 @@ export class MemoryStore {
   private memoryFullError(target: "memory" | "user" | "failure", contentLength: number): MemoryResult {
     const current = this.charCount(target);
     const limit = this.charLimit(target);
+    const entries = this.entriesFor(target).map((raw) => this.decodeEntry(raw).text);
     return {
       success: false,
-      error: `Memory at ${current}/${limit} chars. Adding this entry (${contentLength} chars) would exceed the limit. Replace or remove existing entries first.`,
+      error: `Memory at ${current}/${limit} chars. Adding this entry (${contentLength} chars) would exceed the limit. Replace or remove existing entries first (see the entries list below), then retry this add — all in this turn.`,
+      target,
+      usage: `${current}/${limit} chars`,
+      entry_count: entries.length,
+      entries,
     };
   }
 
--- a/tests/store/memory-store.test.ts
+++ b/tests/store/memory-store.test.ts
@@ -366,6 +366,67 @@ describe("MemoryStore", { concurrency: 1 }, () => {
       assert.ok(raw.includes(existing));
     });
 
+    it("includes current entries in memoryFullError response under reject strategy", async () => {
+      const store = new MemoryStore(makeConfig({
+        memoryCharLimit: 140,
+        memoryOverflowStrategy: "reject",
+      }));
+      await store.loadFromDisk();
+
+      const first = `${TEST_MARKER} first`;
+      const second = `${TEST_MARKER} second`;
+      assert.ok((await store.add("memory", first)).success);
+      assert.ok((await store.add("memory", second)).success);
+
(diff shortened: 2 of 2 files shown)


chandra447:
Review verdict: this is a useful, narrow improvement and the current CI is green (Type Check, Min Supported SDK, Tests, Lint). The existing MemoryResult fields and memory-tool JSON failure path should expose the entries to the agent without a tool-layer change. I recommend one regression test before merge: fill a reject-mode store (and the fifo-evict-cannot-fit path), assert target/usage/entry_count/decoded entries are returned, and assert metadata comments are not leaked. This improves same-turn curation and complements issue #177, but it does not coordinate or suppress a racing background consolidation; that requires a separate fix.

jingyulong:
Hi @chandra447, thanks for the review and the clear guidance.

I have added the regression test you suggested in commit 7dfb924. It lives in tests/store/memory-store.test.ts and covers both overflow paths that populate MemoryResult in memoryFullError:

1. **reject strategy** - fills the store (char limit 140) with two entries, then adds an oversized entry. Asserts success === false, error contains the new guidance text, target === "memory", usage matches ^\d+/140 chars$, entry_count === 2, and entries deep-equals the decoded text. Also asserts no entry contains <!-- or created= so the metadata comment cannot leak.

2. **fifo-evict cannot-fit** - adds one entry (char limit 80), then an oversized entry that alone exceeds the limit (the encoded.length > limit branch in fifoEvictAndAdd). Asserts the same four fields plus entry_count === 1, no metadata leakage, and that the existing entry remains on disk (nothing was rotated).

I also rebased onto the latest main (now includes #179). There were no conflicts - #179 modifies the _add/addWithConsolidation grace-window logic while this PR only touches memoryFullError, so they compose cleanly. Both new tests pass alongside #179 grace-window suite (91 tests, 0 failures locally). The reject and fifo-evict paths bypass addWithConsolidation auto-consolidate branch, so they are unaffected by the grace window and pass on both the pre- and post-#179 tree.

As you noted, this PR improves same-turn curation by returning entries but does not address the background-consolidation coordination race from #177 - that is covered separately by #179. Happy to adjust the test naming or add anything else you would like.

Issue #183, opened 2026-08-15
Critical: sessions.db bloat causing startup hangs and unresponsive agent


srcKod:
## Issue Summary

The `sessions.db` SQLite database in pi-hermes-memory grows without bound, reaching **5.4 GB** in my environment and causing:
- **Startup hangs** — pi becomes unresponsive for minutes
- **Failed operations** — DB queries timeout or hit SQLite limits
- **System instability** — memory pressure from reading large blobs

## Root Cause

Assistant messages containing **tool_result blocks** store the full content of large outputs (file dumps, command outputs, read results) verbatim in the `messages.content` column.

Example: A single assistant message that ran `read` on a large source file stored 1.2 MB of file content. My DB has **695 messages exceeding 1 MB each** (~2 GB total from this pattern alone).

The issue is in two places:
1. `session-parser.ts` — `extractTextContent()` concatenates tool_result blocks into the content string
2. `session-indexer.ts` — `parseMessageEntry()` passes raw content to DB without any truncation

## Code Locations

- `src/store/session-parser.ts` — `extractTextContent()` iterates content blocks and includes tool_result content
- `src/store/session-indexer.ts` — `parseMessageEntry()` returns `{content, ...}` without length check

### 1. Skip tool_result blocks in extractTextContent()

In `session-parser.ts`, tool_result blocks should be skipped entirely (not concatenated):

```typescript
function extractTextContent(content: unknown): string {
  if (!Array.isArray(content)) return String(content ?? '');
  const parts: string[] = [];
  for (const block of content) {
    if (block.type === 'text' && typeof block.text === 'string') {
      parts.push(block.text);
    }
    // Skip tool_result blocks - they contain ephemeral data (file contents, command outputs)
    // that bloats the DB. Tool names are already tracked in tool_calls.
  }
  return parts.join('');
}
```

### 2. Add content length cap as safety net

In `constants.ts`, add:
```typescript
export const DEFAULT_MAX_MESSAGE_CONTENT_LENGTH = 100 * 1024; // 100 KB
```

In `session-indexer.ts`, truncate before DB insert:
```typescript
const MAX = DEFAULT_MAX_MESSAGE_CONTENT_LENGTH;
const truncatedContent = content.length > MAX
  ? content.slice(0, MAX/2) + '\n... (truncated, ' + content.length + ' chars total)' + content.slice(-MAX/2)
  : content;
```

## Impact of Fix

[...]

PR #184, opened 2026-08-16
feat: FTS5 query quality and session retention pruning

srcKod:
## Summary

Narrows this PR to the value that is **unique** versus the already-merged bloat fix (#187). It drops the truncation/parser changes that #187 already covers with its own `100 * 1024` cap, and adds the split-out query-quality and retention work the previous review requested.

### 1. FTS5 query quality (session + memory search)
- **Stop-word filtering**: drop ~100 common English filler words (the, how, to, for, in, ...) from natural-language queries so they can't dominate FTS5 ranking or cause misleading AND-style misses. Explicit operators (`AND`/`OR`/`NOT`/`NEAR`) and quoted phrases are preserved untouched.
- **Broader FTS5 error detection**: `isFts5QueryError` now also recognizes `malformed` and `sql logic error`.
- **Graceful memory reconciliation**: when the FTS5 index is broken, `reconcileMarkdownMemoryScope` returns a graceful no-op (entry still persists to Markdown) instead of surfacing a scary `database disk image is malformed` error; the memory tool shows a friendly re-index hint.

### 2. Session retention pruning (bounds DB growth — complements #187 and #186)
- New `sessionRetentionDays` config (default `180`) with `DEFAULT_SESSION_RETENTION_DAYS`.
- On startup, after persistence initializes, prune sessions older than the window and their messages. Counts messages upfront and deletes messages before sessions (the `messages` table is not `ON DELETE CASCADE`) with accurate per-table removed counts.
- This bounds the session index database over time, complementing #187 (unbounded growth prevention) and #186 (deferring startup integrity scans for large DBs) — no file overlap with either.

## Test notes (Windows)
All new regression tests pass (`fts-query.test.ts`, and the added `pruneOldSessions` cases in `session-indexer.test.ts`); `tsc --noEmit` is clean. The pre-existing failures in `extension-root-migration.test.ts`, `sync-markdown-memories.test.ts`, `db.test.ts`, `sqlite-native.test.ts`, and others are **Windows-only environment issues** (EPERM/EBUSY on locked WAL/temp files) that reproduce identically on standalone `origin/main` and are unrelated to these changes.

## Related
- Closes/depends on discussion in #183 (sessions.db bloat / startup hangs).
- Complements #186 (open) — separate files, no conflict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Opt-in memory lifecycle timings**: `PI_TIMING=1` now reports startup synchronization and loading, backfill and live-index callbacks, shutdown flush/index/wait/close, and database open/integrity-check/checkpoint durations. Normal runs produce no timing output.
 
+- **Opt-in SQLite session retention** ([#183](https://github.com/chandra447/pi-hermes-memory/issues/183)): new `sessionRetentionDays` setting (default `0`, disabled). When set to a positive value, sessions whose JSONL source file has not been updated within the window are pruned from SQLite at startup — rows only, the JSONL files themselves are never deleted — and both the deferred backfill and `/memory-index-sessions` skip files outside the window, so pruned sessions stay pruned instead of being re-indexed on every startup. With the default `0`, pruning is fully disabled and startup keeps the legacy count-only backfill preflight.
+
+### Changed
+
+- **Natural-language search now filters common stop words** ([#184](https://github.com/chandra447/pi-hermes-memory/pull/184)): session and memory searches drop function words (`the`, `is`, `what`, …) before FTS5 matching, improving precision for natural-language queries. Intent-bearing words (`can`, `need`, `off`, `like`, `get`, …) are deliberately kept. A query made up entirely of stop words no longer returns an empty result — it degrades to a scoped literal substring search over the original terms.
+
+### Fixed
+
+- **FTS5 index errors during markdown sync are surfaced instead of hidden** ([#184](https://github.com/chandra447/pi-hermes-memory/pull/184)): when the FTS5 search index is broken, memory writes still succeed in Markdown, but the tool result and `/memory-sync-markdown` now report the degraded search state together with the repair command instead of showing a silent zero-count success. Genuine database corruption is unaffected by the fallback and now recovers through the markdown sync path too: reconciliation runs inside `DatabaseManager.withCorruptionRecovery()`, so a corrupt `sessions.db` is quarantined, rebuilt, and the sync retried instead of failing every memory write.
+
 ### Fixed
 
 - **Clean process exit on Node 24+** ([#193](https://github.com/chandra447/pi-hermes-memory/issues/193), reported by [@vanneswong](https://github.com/vanneswong)): `better-sqlite3` 12.x inherits Node's raw `node::ObjectWrap`, whose destructor calls `RemoveEnvironmentCleanupHook` after Node has torn down its Environment — so on Node 24 (observed on Linux aarch64) Pi could abort with SIGABRT and a non-zero exit code at every shutdown, after all work completed. The dependency moves to `better-sqlite3@^13.0.3`, the first N-API/node-addon-api release, which removes that crash path. The public API this extension uses is unchanged; the declared engines (`>=22`) cover every Pi-supported runtime.
--- a/README.md
+++ b/README.md
@@ -495,6 +495,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
   "memoryDir": "~/.pi/agent/pi-hermes-memory",
   "projectsMemoryDir": "projects-memory",
   "sessionSearch": { "variant": "legacy" },
+  "sessionRetentionDays": 0,
   "llmModelOverride": "openrouter/deepseek/deepseek-v4-flash",
   "llmThinkingOverride": "off",
   "childExtensionPaths": ["~/.pi/agent/git/github.com/example/custom-provider-extension/index.ts"],
@@ -532,6 +533,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `memoryDir` | `~/.pi/agent/pi-hermes-memory` | Custom directory for extension storage files |
 | `projectsMemoryDir` | `projects-memory` | Subdirectory under `~/.pi/agent/` for project-scoped memory |
 | `sessionSearch` | `{ "variant": "legacy" }` | Session search implementation: `legacy` keeps the existing SQLite/FTS snippet search; `anchors` uses the opt-in Markdown request surface and returns compact JSONL line-range anchors from `~/.pi/agent/sessions/` |
+| `sessionRetentionDays` | `0` | Opt-in SQLite session retention, in days. `0` (default) disables pruning entirely and keeps the legacy count-only backfill preflight. When positive, sessions whose JSONL source file was last modified longer ago than the window are pruned from SQLite at startup — **rows only; the JSONL files in `~/.pi/agent/sessions/` are never deleted** — and both the deferred backfill and `/memory-index-sessions` skip files outside the window, so pruned sessions stay pruned instead of being re-indexed |
 | `llmModelOverride` | unset | Optional model override for background review (direct and subprocess), correction save, session flush, and consolidation |
 | `llmThinkingOverride` | unset | Optional thinking override for those LLM calls; valid values are `off`, `minimal`, `low`, `medium`, `high`, and `xhigh`. If `llmModelOverride` is set and this is omitted, review/child calls default to `off` |
 | `childExtensionPaths` | unset | Trusted provider/auth extension sources explicitly allowed in isolated child Pi processes. Values are passed to Pi's standard `-e` resolver, so absolute paths, `~/...`, paths relative to the child working directory, and `git:`/`npm:` package sources are supported. Sibling packages matching the `*-oauth-adapter`/`*-auth-adapter` naming convention (including scoped packages, via their `package.json` `pi.extensions` manifest) are detected automatically. This setting is only needed for custom providers or adapters that are not detected. In-process direct transport (the default for review/flush/correction/consolidation) doesn't need it, since it reads whatever provider auth is already registered. |
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -294,7 +294,7 @@ Agent has access to tools:
 - [ ] Auto-index session on shutdown
 - [ ] Char limits: MEMORY.md 2,200 → 5,000, USER.md 1,375 → 5,000, project 2,200 → 5,000
 - [ ] Config: `sessionSearchEnabled: boolean` (default: true)
-- [ ] Config: `sessionRetentionDays: number` (default: 90)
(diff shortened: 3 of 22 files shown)


chandra447 (changes requested):
Thanks for tackling #183. I cannot merge this as written because it does not address the primary bloat source: `tool_result` blocks are still included by session parsing. In addition, `indexSessionOnce` ignores the supplied `maxMessageLength` option and always uses the constant, the truncation result can exceed the advertised limit because the marker is appended after slicing, and `pruneOldSessions()` reports `messagesRemoved` as the number of deleted sessions. Please split unrelated FTS/backfill changes from the bloat fix, skip tool results in both parser paths, and add regression tests for those cases.

srcKod:
Rebasing on the current main after #187 merged, I narrowed this PR to the parts that are still unique:
  - FTS5 stop-word filtering for natural-language session/memory queries (explicit operators and quoted phrases are preserved untouched)
  - Session retention pruning via a new sessionRetentionDays config (default 180) — bounds DB growth over time
  - Graceful degradation in reconcileMarkdownMemoryScope when the FTS5 index is broken, plus a friendlier memory-tool hint
  
  I dropped the truncation/backfill edits here since #187 already caps content at its own limit. #186 (deferred startup integrity scans) is also complementary — separate file (db.ts), no overlap.
 
  Note: the remaining test failures in this PR's CI on Windows (sync-markdown-memories, db.test, sqlite-native, migration tests) reproduce identically on clean main — they're pre-existing EPERM/EBUSY file-lock issues on Windows, not from these changes.

chandra447 (changes requested):
Thanks for narrowing this after #187/#186; the scope is much clearer and the CI is green. I still cannot approve this revision because two changes introduce release-blocking behavior:

1. The default startup prune is not compatible with the existing incremental backfill contract. Deleting an old `sessions` row cascades its `session_files` metadata, while `needsBackfill()` compares all JSONL files with indexed session rows and `indexChangedSessions()` treats missing metadata as changed. As long as historical JSONL files remain on disk, every startup schedules a backfill and re-indexes pruned sessions (up to the 50-file cap), so retention is neither durable nor cheap. Please make retention/backfill agree on the eligible file set (or retain metadata/tombstones), and add an end-to-end regression proving old on-disk files are neither reindexed nor repeatedly schedule startup backfills.

2. `isFts5QueryError()` now categorizes `database disk image is malformed` and generic `SQL logic error` as recoverable FTS errors. `reconcileMarkdownMemoryScope()` catches those and returns a no-op, which suppresses `DatabaseManager` corruption recovery rather than letting the corrupt database be quarantined/rebuilt. The former is explicitly a corruption signal used by #186, and the latter is too broad to safely swallow. Please keep corruption errors on the recovery path; only handle a narrowly identified FTS query/index error, with an integration test that proves corrupted DB recovery still runs.

Also please document and test the new configuration semantics (including whether retention can be disabled). The current default silently deletes existing searchable session history after 180 days, so that policy needs an explicit, backwards-compatible design.

srcKod:
Addressed the review items:

 1. Retention ↔ backfill alignment
 Threaded retentionCutoffMs through scheduleSessionBackfill → needsBackfill() / indexChangedSessions(). Expired JSONL files (outside the retention window) are now skipped by both, so pruned sessions are never re-indexed and never re-trigger startup backfills. pruneOldSessions keys on file mtime_ms (with started_at fallback) so retention and backfill agree on the same eligible file set. Added 4 integration tests proving old files are neither re-indexed nor schedule a backfill.

 2. Opt-in retention default
Changed DEFAULT_SESSION_RETENTION_DAYS from 180 → 0 (disabled). Users explicitly opt in with a positive value; 0 disables even after opting in. Config parsing ignores negative/non-numeric values. Updated docs in types.ts and constants.ts.

 3. FTS5 error handling fix
 The original isFts5QueryError broadening ("malformed" / "sql logic error") was too broad and could mask genuine persistence failures during markdown writes. Fixed by:
 - Splitting reconcileMarkdownMemoryScope into two steps: 
     - persist markdown entries — errors propagate
     - orphan deletion/FTS sync — only swallows actual FTS5 errors
 - Added console.warn with actionable message when FTS5 error is caught
 - memory-tool.ts now uses shared isFts5QueryError and shows the actual error + fix instruction (/memory-sync-markdown)

 All affected tests pass (78/78 retention/backfill, 75/75 FTS5/memory, 90/94 memory search), TypeScript clean.

chandra447:
Thanks for narrowing this to query quality, retention, and the FTS failure path. The changes are meaningful, but the current head still has two release blockers. `reconcileMarkdownMemoryScope()` catches generic `malformed` and `SQL logic error` failures and returns a zero-count success after the SQLite transaction rolls back. That hides database corruption and prevents the caller or `/memory-sync-markdown` from reporting or repairing the failed index. Restrict the classifier to genuine FTS query errors and preserve database recovery. The all-expired, no-backfill-timestamp startup path also still schedules an empty backfill because `needsBackfill()` falls through to its periodic timestamp check. Add a regression and return no work when no retained files remain.

srcKod:
Both release blockers are addressed, and the branch is rebased onto current main (`5f6fa99`, includes #205/#204/#203). New head: `99a7fbe`.

**1. FTS5 classifier restricted to genuine query errors (corruption recovery preserved)**

- `isFts5QueryError()` no longer matches `database disk image is malformed` or generic `SQL logic error`. It only matches genuinely identified FTS5 query/index failures (`no such column: rank`, `syntax error in FTS5 query`, `unterminated string`).
- `reconcileMarkdownMemoryScope()` now runs its persistence inside `DatabaseManager.withCorruptionRecovery()`, so a corrupt DB is quarantined, rebuilt, and the operation is retried on the healthy database — the #186 corruption signal is on the recovery path again and can no longer be swallowed into a zero-count success.
- Only the narrow FTS5 error set gets the soft fallback, and even then it logs a `console.warn` pointing at `/memory-sync-markdown` so a broken FTS index is visible and repairable.
- New integration test `tests/integration/corruption-recovery-integration.test.ts` ("Corruption recovery during markdown sync") proves end-to-end that a corrupt DB is quarantined + rebuilt + re-persisted through `reconcileMarkdownMemoryScope`, and that a genuine FTS5 query error still gets the soft fallback. Passing.

**2. No empty backfill when all files are expired**

- `needsBackfill()` no longer falls through to the periodic timestamp check when retention is enabled and no file remains inside the window: it returns no work instead of scheduling an empty backfill on every startup.
- Three regression tests added in `tests/store/session-indexer.test.ts`:
  - all-expired + no backfill timestamp → `false` (was `true` before the fix)
  - retained file exists + no timestamp → still `true` (real work is still scheduled)
  - retention disabled → periodic check behaves exactly as before

**3. Rebase onto current main**

Conflicts resolved in `src/index.ts` (session_start now keeps #204's `runRecoveryMaintenance` snapshot sweep and `pruneEphemeralReviewSessions`, with retention pruning running in the same `persistenceInitialized` block *before* `scheduleSessionBackfill`, so pruned sessions are never re-indexed by a backfill started in the same startup) and `src/handlers/session-backfill.ts` (kept #200's `measureLifecycleSync` instrumentation while threading `retentionCutoffMs` through `needsBackfill` / `indexChangedSessions`).

**Verification**

- TypeScript clean (`tsc --noEmit`).
- Full suite on the rebased branch: 819 tests, 753 pass, 56 fail.
- The 56 failures are the exact same set as a clean `upstream/main` run on this Windows machine (file-lock/EBUSY issues in db recovery, atomic locks, migration, sqlite-native) — verified by running the identical suite in a clean worktree of `5f6fa99` and diffing the failure lists: zero regressions from this branch.

chandra447 (changes requested):
Thank you for using pi-hermes-memory and for contributing the FTS5 query-quality, session-retention, and corruption-recovery work in this PR. The retention and backfill alignment is useful, and the recovery wrapper is the right direction. I am requesting changes before merge for the following concrete issues.

1. Retention is bypassed by the manual reindex command.

`src/handlers/index-sessions.ts:37-42` still calls `indexAllSessions()` without the retention cutoff. `pruneOldSessions()` removes rows but leaves JSONL files on disk, so `/memory-index-sessions` re-imports every expired file and resurrects the rows that retention just removed. Please thread the cutoff through this path, or document and enforce that manual indexing intentionally bypasses retention.

2. The default startup preflight now stats every session file twice when counts differ.

`needsBackfill()` performs a metadata scan even when `retentionCutoffMs` is `0`, then `indexChangedSessions()` scans the same files again. This replaces the old cheap count check with synchronous `fs.statSync` work before the deferred backfill. On a large session directory this can reintroduce the startup latency this area is meant to avoid. Preserve the immediate count-only path when retention is disabled and avoid duplicate metadata work when it is enabled.

3. The FTS5 fallback prevents the memory tool's repair message from being shown.

`reconcileMarkdownMemoryScope()` catches the FTS error and returns zero counters at `src/store/sqlite-memory-store.ts:500-515`. `src/tools/memory-tool.ts:254-258` therefore sees a successful reconciliation and cannot return its `/memory-sync-markdown` guidance. Keep the Markdown write successful, but represent the degraded SQLite result explicitly so the caller can show the user that search may be stale and how to repair it.

4. Stop-word filtering silently changes real searches.

An all-stop-word query such as `what is it` normalizes to an empty string, and both search entry points return `[]` before their existing fallback paths. The list also removes terms that can be meaningful, including `can`, `need`, `off`, `like`, and `get`. Please either retain a small, safer function-word set and provide a raw/LIKE fallback when filtering removes every term, or document and test the intentional zero-result contract.

5. One new test is tautological, and the classifier change is not actually present.

The assertion in `tests/store/fts-query.test.ts` contains `|| true`, so it cannot fail. Also, comparing `isFts5QueryError()` with `main` shows the implementation is logically unchanged. The actual behavior change is the new `withCorruptionRecovery()` wrapper around reconciliation. Please replace the tautology and correct the comments and integration-test narrative so they describe the behavior that changed.

6. Document the new destructive-capable setting.

`sessionRetentionDays` is absent from the README configuration table and changelog. Please document the default `0`, the opt-in behavior, the file-mtime basis, and that pruning removes SQLite rows but does not delete JSONL files.

I checked the Pi extension contracts as well. `session_start`, the existing handler wiring, and the SDK usage in this PR are valid. Both local targeted verification and GitHub CI pass, including Type Check, Min Supported SDK, Tests, and Lint. The requested changes are about retention/search behavior and verification accuracy, not Pi API signatures.

srcKod:
One more change since the last push: the aggregate result from `reconcileMarkdownFailureScopes()` was keeping the **last** degraded reason across failure scopes instead of the first — the comment already said 'first', the code was just overwriting on every iteration. Guarded with `!total.degraded` so the first scope's reason is the one surfaced to the user. Regression test added (`tests/store/sqlite-memory-store.test.ts` — two projects with different FTS5 error messages; aggregate matches the first, not the second).

Branch is now rebased onto `671a27f` (v0.9.7) and force-pushed. New head: `b9224ea`. 6 commits ahead of `origin/main`, TypeScript clean, 188/188 affected-module tests pass.

chandra447 (commented):
Adversarial review (4 reviewers) — no merge-blocking defects found. Coherent scoping vs #187. One residual question for the author: the Windows-only test failures (extension-root-migration, sync-markdown-memories, db, sqlite-native) are claimed pre-existing/environmental — worth linking the reproducing main-branch run so future reviewers don't relitigate. Otherwise good to merge.

PR #186, opened 2026-08-16
fix: skip full-DB integrity scans at startup

zenworr:
Startup cost grows linearly with the extension DB (~8 s at 1.1 GiB, paid twice per session): openUnchecked ran two full-DB PRAGMA quick_check scans on every process start.

Open without the scans and schedule a single deferred quick_check on the next event-loop turn (once per open). A failure funnels into the existing quarantine-and-rebuild recovery, so a recoverably-corrupt file is still repaired on open (getLastRecovery() reflects it) without adding scan time to the startup path.

### DIFF
--- a/src/store/db.ts
+++ b/src/store/db.ts
@@ -162,6 +162,7 @@ export class DatabaseManager {
   private readonly recoveryOptions: ResolvedDatabaseRecoveryOptions;
   private lastRecovery: DatabaseRecoveryResult | null = null;
   private openGuard: (() => void) | null = null;
+  private pendingOpenIntegrityScan: Promise<void> | null = null;
   private activeRecoveryLease: { coordinator: AtomicLockCoordinator; key: string; token: string } | null = null;
 
   constructor(memoryDir: string, recoveryOptions: DatabaseRecoveryOptions = {}) {
@@ -267,8 +268,9 @@ export class DatabaseManager {
       fs.mkdirSync(dir, { recursive: true });
     }
 
+    let opened: DatabaseLike;
     try {
-      return this.openUnchecked();
+      opened = this.openUnchecked();
     } catch (err) {
       if (!DatabaseManager.isCorruptionError(err)) {
         throw err;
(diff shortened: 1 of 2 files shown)


chandra447 (changes requested):
Thanks for the startup-performance improvement. I reproduced a regression locally: `npm run check && npm test` fails `tests/store/db.test.ts` — “repairs recoverable corruption on open and preserves readable rows”. Without `quick_check`, opening the recoverably-corrupt fixture succeeds and no rebuild occurs (`getLastRecovery()` is undefined). Please preserve the corruption-on-open recovery guarantee, add/update coverage for the faster path, then this can be reconsidered.

zenworr:
Reworked in response to the review: the corruption-on-open guarantee is preserved — the scan just no longer blocks startup. openUnchecked opens without the two full-DB quick_check scans; a single quick_check now runs deferred on the next event-loop turn (once per open) and failures funnel into the same quarantine-and-rebuild recovery. Opening a recoverably-corrupt fixture is therefore still followed by a rebuild (getLastRecovery = rebuilt, rows salvaged, file quarantined), just not synchronously inside open. Startup cost is constant regardless of DB size.

chandra447 (changes requested):
Thanks for the revised deferred scan and for restoring the corruption-recovery coverage. The current head passes `npm run check && npm test` locally, but I found a reopen-path regression that keeps this out of the release for now. `pendingOpenIntegrityScan` is set once and never cleared, so after `getDb() -> close() -> getDb()` the second open skips `scheduleOpenIntegrityScan()` permanently. I reproduced this by stubbing `assertIntegrityOk` after the first scan: `deferred_scans_after_reopen=0`. Please clear/reset the pending promise after the scan settles (and when appropriate during close), and add a regression test proving a reopened manager schedules and completes a new integrity scan.

zenworr:
The reopen path now reschedules the scan: pendingOpenIntegrityScan is cleared when the scan settles (identity-guarded so a stale timer cannot clear a newer scan) and in close(), so getDb() → close() → getDb() schedules a fresh scan. Added a regression test that reopens the same manager on a recoverably-corrupt index page and asserts the rebuild.

chandra447 (approved):
Re-reviewed the latest head. The pending scan state is now cleared after settlement and on close, the new reopen/corruption regression passes, and I independently reproduced a fresh deferred scan after reopen (`deferred_scans_after_reopen=1`). Local type-check and the full suite passed; CI is green. Approved for the 0.9.6 release.

PR #189, opened 2026-08-17
fix: preserve child extension source specs

xz-dev:
## Summary

- pass configured `childExtensionPaths` values unchanged to Pi's standard `-e` resolver
- support portable `~/...`, relative, `git:...`, and `npm:...` extension sources
- retain local existence checks for Hermes-discovered auth adapter paths

## Problem

Hermes currently calls cwd-relative `path.resolve()` and `existsSync()` on every configured child extension value before invoking Pi. That silently drops Pi-native package sources and home-relative paths, forcing machine-specific absolute paths for custom providers in isolated subprocesses.

Pi already owns this resolution contract: CLI `-e` values are passed through `resolveExtensionSources()`, which handles local paths and package sources. This change leaves configured values intact so the child Pi process uses that standard behavior while preserving the existing `--no-extensions` isolation boundary.

## Verification

- `npm run check`
- `npm run check:min-sdk` (Pi 0.80.1)
- `npm test` (all 47 test files passed)
- real Pi smoke test with a `~/.pi/agent/...` provider extension source
- remote cross-home validation: Dotdrop rendered the same config to `/root/...`, and isolated Hermes + provider loading resolved the expected models without `No models match` warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


chandra447:
Thanks for raising this and for leveraging Pi's extension system so thoughtfully.

This is the right direction: the isolated child still starts with `--no-extensions`, while configured trusted sources are now passed directly to Pi's standard `-e` resolver. That restores support for home-relative, child-CWD-relative, npm, and git sources without Hermes duplicating or narrowing Pi's resolution contract.

The locally discovered Hermes/auth-adapter paths retain their existence checks, which preserves the existing isolation and safety boundary. I reviewed the change and found no issues. Thank you!

PR #191, opened 2026-08-21
fix(review): support header-only OAuth auth

xz-dev:
## Summary

- accept request authentication supplied through provider headers, not only `apiKey`, in direct memory completions
- recognize only credential-bearing headers (`Authorization`, `x-api-key`, and `cf-aig-authorization`) while forwarding all resolved headers unchanged
- normalize thrown 401/403 failures and assistant `stopReason: "error"` responses through the same one-retry auth-rotation path
- compare header names case-insensitively without collapsing duplicate-case entries, and tolerate `ProviderHeaders` null values
- retain Pi's SDK-derived auth types and public `modelRegistry.getApiKeyAndHeaders()` API

## Root cause

Pi OAuth-backed providers can resolve request auth entirely through headers. The direct review path originally required `apiKey`, and the first PR revision still assumed provider auth failures were thrown. In production, `completeSimple()` normally returns an assistant message with `stopReason: "error"` and `errorMessage`; static headers could also be mistaken for credentials.

The updated path accepts header-owned auth, distinguishes credential headers from ordinary provider headers, and re-resolves effective request auth once after either form of 401/403 failure. If the credentials did not change, or the retry also fails, it returns `provider_error` without a third completion.

## Verification

- `node --test --import tsx tests/handlers/review-memory-ops.test.ts` — 31/31 passed
- `npm run check`
- `npm run check:min-sdk` — source type-checks against Pi SDK 0.80.6
- `git diff --check`
- regressions cover error assistant responses, `User-Agent`-only auth, null credential values, casing-only changes, duplicate-case header entries, and the strict one-retry ceiling

### DIFF
--- a/src/handlers/review-memory-ops.ts
+++ b/src/handlers/review-memory-ops.ts
@@ -92,7 +92,7 @@ type ReviewModelRegistry = ExtensionContext["modelRegistry"];
 /** Derived from the installed SDK so headers/baseUrl track ProviderHeaders instead of a local mirror. */
 export type ResolvedRequestAuth = Awaited<ReturnType<ReviewModelRegistry["getApiKeyAndHeaders"]>>;
 
-type DirectReviewAuth = Omit<Extract<ResolvedRequestAuth, { ok: true }>, "ok"> & { apiKey: string };
+type DirectReviewAuth = Omit<Extract<ResolvedRequestAuth, { ok: true }>, "ok">;
 
 export function buildDirectReviewCompletionOptions(
   model: Model<Api>,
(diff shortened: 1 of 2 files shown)


chandra447:
Thank you for using pi-hermes-memory and for contributing header-only OAuth support. This is a useful compatibility improvement, and the basic forwarding of `apiKey`, `headers`, and `env` through Pi's public `modelRegistry.getApiKeyAndHeaders()` path is the right foundation. I also appreciate the focused tests and green CI.

I found two remaining blockers before merge:

1. The rotation path currently expects `completeSimple()` to throw on a 401/403. Pi's real provider path normally resolves an assistant response with `stopReason: "error"` and the provider failure in `errorMessage`, so the nested `catch` is not reached. A stale OAuth header therefore is not re-resolved and retried in production even though the throwing test passes. Please normalize both thrown failures and error assistant responses through the same one-retry auth-rotation path, and add a production-shaped test that returns `{ stopReason: "error", errorMessage: "...401..." }`.

2. `hasRequestAuth()` currently treats any non-empty header as credentials. Pi's resolved headers can also include ordinary static headers such as `User-Agent`; those should not turn a logged-out model into an authenticated one. Please check credential-bearing headers case-insensitively (`Authorization`, `x-api-key`, and `cf-aig-authorization`, alongside `apiKey`) and add a `User-Agent`-only regression. It would also be good to compare normalized header names so casing-only changes do not look like credential rotation.

With those changes, this should remain a narrow provider-compatibility fix using Pi's public constructs rather than private auth internals. Thanks again for using the extension and helping make it work reliably across OAuth-backed providers.

chandra447:
Thanks for adding header-only OAuth support through Pi's public `modelRegistry.getApiKeyAndHeaders()` API. This is useful provider compatibility work. I am holding the current revision for two production cases. `completeSimple()` normally returns an assistant message with `stopReason: "error"` and `errorMessage`, so a 401 or 403 in that response must enter the same one-retry auth-rotation path as a thrown error. Also, `hasRequestAuth()` must inspect only credential-bearing headers, not static headers such as `User-Agent`, and must tolerate the latest `ProviderHeaders` values that can be null. Normalize header names before comparing rotated credentials. Please include regressions for an error response, a User-Agent-only response, and null header values.

xz-dev:
Addressed both blockers and rebased this branch onto the current `main`.

Changes:
- normalized thrown auth failures and assistant `{ stopReason: "error", errorMessage }` 401/403 responses through the same single-retry rotation path;
- limited authentication detection to `apiKey` or case-insensitive `Authorization`, `x-api-key`, and `cf-aig-authorization` headers;
- handled `ProviderHeaders` null values while forwarding all resolved headers unchanged;
- normalized header names for credential comparison without collapsing duplicate-case entries;
- retained the SDK-derived `getApiKeyAndHeaders()` result type from #209.

Added regressions for the production-shaped error response, `User-Agent`-only headers, null credential values, casing-only changes, duplicate-case entries, and a strict two-completion ceiling even when a third distinct credential is available.

Validation on `a47815e`:
- focused review-memory-ops tests: 31/31
- full suite: all 50 test files
- type check
- minimum SDK 0.80.6 check
- `git diff --check`

chandra447 (commented):
Adversarial review (4 reviewers) — no blockers found. Header allowlist scoping, case-insensitive compare without collapsing duplicates, null tolerance, and the strict one-retry ceiling all hold up. Clean to merge once #192 lands (both touch adjacent auth paths; rebase to be safe).

PR #192, opened 2026-08-21
fix(review): scope background cancellation to session

xz-dev:
## Summary

- decouple fire-and-forget memory review from the active interactive turn's abort signal
- give background review one session-scoped `AbortController` shared by direct and subprocess transports
- abort and boundedly await active review from an idempotent `session_shutdown` handler
- suppress fallback, stale memory writes, and stale UI notifications after session cancellation
- guard direct review before provider work and again at the memory mutation boundary

## Root cause

Background review deliberately outlives the interactive turn that triggered it, but its subprocess fallback forwarded `ctx.signal`. Aborting that turn therefore cancelled unrelated review work. Simply removing the signal fixed turn coupling but also removed session ownership, allowing old review work to survive reload, session replacement, fork, or shutdown.

The updated implementation uses a dedicated signal for the lifetime of the session. Shutdown aborts it and waits up to a local one-second cleanup grace; normal review execution retains its independent 120-second timeout. Cancellation is also rechecked after acquiring the Markdown mutation lock so an old session cannot begin a queued memory write after shutdown.

## Verification

- focused background-review, direct-completion, and MemoryStore tests passed
- `npm run check`
- `npm run check:min-sdk` — source type-checks against Pi SDK 0.80.6
- `git diff --check`
- regressions prove turn abort is ignored, session shutdown cancels and waits, shutdown remains bounded if a provider ignores abort, cancelled work stays silent, fallback does not start, and a queued memory mutation cannot write after cancellation

### DIFF
--- a/src/handlers/background-review.ts
+++ b/src/handlers/background-review.ts
@@ -122,7 +122,6 @@ const SESSION_REVIEW_SHUTDOWN_GRACE_MS = 1000;
 function awaitUpTo(promise: Promise<void>, ms: number): Promise<void> {
   return new Promise((resolve) => {
     const timer = setTimeout(resolve, ms);
-    timer.unref();
     promise.then(
       () => {
         clearTimeout(timer);
--- a/src/handlers/background-review.ts
+++ b/src/handlers/background-review.ts
@@ -31,9 +31,11 @@ export interface BackgroundReviewDeps {
   runDirectReview?: typeof runDirectMemoryCompletion;
   execChildPrompt?: typeof execChildPrompt;
   /** Test-only hook: called once runReview() has fully settled (after the
-   * fire-and-forget review work completes and reviewInProgress resets),
+   * fire-and-forget review work completes and activeReview clears),
    * since production callers never await runReview() directly. */
   onReviewSettled?: () => void;
+  /** Test-only override for session_shutdown's wait bound. */
+  shutdownGraceMs?: number;
 }
 
 export interface ReviewPromptInput {
@@ -113,17 +115,41 @@ function diagnosticDetail(value: unknown): string {
   return detail.replace(/\s+/g, " ").slice(0, 300);
 }
 
+/** Shutdown wait only. Review still uses its own 120s timeout; 1s covers abort
+ * listener + cancel-file write without hanging Pi's awaited session_shutdown. */
+const SESSION_REVIEW_SHUTDOWN_GRACE_MS = 1000;
+
+function awaitUpTo(promise: Promise<void>, ms: number): Promise<void> {
+  return new Promise((resolve) => {
+    const timer = setTimeout(resolve, ms);
+    timer.unref();
+    promise.then(
+      () => {
+        clearTimeout(timer);
+        resolve();
+      },
+      () => {
+        clearTimeout(timer);
+        resolve();
+      },
+    );
+  });
+}
+
 async function runSubprocessReview(
   pi: ExtensionAPI,
   prompt: string,
   config: MemoryConfig,
   execChild: typeof execChildPrompt,
-  ctx: Pick<ExtensionContext, "cwd" | "model" | "signal">,
+  ctx: Pick<ExtensionContext, "cwd" | "model">,
+  signal: AbortSignal,
 ): Promise<{ code: number; stdout?: string; stderr?: string }> {
   return execChild(pi, prompt, config, {
     cwd: ctx.cwd,
     model: resolveChildPiModel(ctx.model),
-    signal: ctx.signal,
+    // Session-scoped signal only. The turn signal belongs to the interactive
+    // agent run; forwarding it cancels unrelated review on a later user abort.
+    signal,
     timeoutMs: 120000,
   });
 }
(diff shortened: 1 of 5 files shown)


chandra447:
Thank you for using pi-hermes-memory and for taking the time to improve the background-review fallback. The core observation here is good: the active turn's abort signal should not automatically own unrelated fire-and-forget review work, and the focused coverage plus green CI are appreciated.

I am holding the merge for one lifecycle issue. Passing `signal: undefined` completely removes session ownership as well as turn ownership. During `/reload`, new/resume/fork, or shutdown, the old child can therefore continue for up to 120 seconds, write review results after a replacement session has started, and notify through a stale context.

Please keep the turn-signal decoupling, but give the background review its own `AbortController`:

- pass that dedicated signal to both the direct review and subprocess fallback;
- abort and boundedly await any active review from an idempotent `session_shutdown` handler;
- extend the regression coverage to prove a turn abort is ignored while session shutdown cancels the child and prevents stale writes/notifications.

That follows Pi's lifecycle contract for session-scoped background work without losing the reliability improvement this PR is aiming for. Thanks again for exercising the extension in a real workflow and contributing the fix.

chandra447:
Thanks for isolating the turn-signal bug and for adding focused regression coverage. I am holding this revision until the background review owns a session-scoped AbortController. The current `signal: undefined` decouples the interactive turn, but it also lets review work survive reload, session replacement, or shutdown and then use stale stores or UI context. Pass a dedicated signal to both transports, abort it from an idempotent `session_shutdown` handler, and boundedly await the active review. Keep the turn-abort regression and add the session-shutdown cancellation case.

kdejaeger:
Thanks for isolating the background-review cancellation issue. I investigated this against the installed `pi-hermes-memory@0.9.7` and Pi `0.84.4`; the `ctx.signal` coupling is reproducible and appears to be a real regression.

## Regression evidence

`runSubprocessReview()` in the installed package has this shape:

```ts
ctx: Pick<ExtensionContext, "cwd" | "model" | "signal">

return execChild(pi, prompt, config, {
  cwd: ctx.cwd,
  model: resolveChildPiModel(ctx.model),
  signal: ctx.signal,
  timeoutMs: 120000,
});
```

Comparing published versions shows:

- v0.9.4 passed `signal: undefined` and did not couple the fallback to the interactive run.
- v0.9.5 introduced active `cwd`/model inheritance and changed this to `signal: ctx.signal`.
- v0.9.6 and v0.9.7 retain that behavior.

Pi's `ExtensionContext.signal` is the active interactive run's `Agent.abortController.signal`. This creates two failure paths:

1. If the user aborts while a fire-and-forget review is running, the fallback receives the same abort.
2. If the parent run is already aborted when `turn_end` is emitted, the review starts with an already-aborted signal.

`execChildPrompt()` responds to that signal by writing the temporary `cancel` marker before spawning the child. The watchdog polls that marker and terminates the child process tree. I verified this directly with the installed implementation: an already-aborted signal produces the cancellation marker before `pi.exec()` is invoked; omitting the signal does not.

The exact observed warning was:

```text
Memory auto-review failed in both transports.
Direct: aborted.
Subprocess: ... child timed out after 120000ms; terminating process tree
```

The standalone subprocess itself is not generally slow: the exact configured model/CLI completed a small prompt in about 3 seconds and a roughly 58 KB review prompt in about 14 seconds, both with exit code 0. The `router/*` model warnings are expected stderr noise because the child uses `--no-extensions` while the global enabled-model list still contains the router profiles.

## Important scope distinction

In v0.9.7 the background direct call does **not** pass `ctx.signal` to `runDirectMemoryCompletion()`. That function creates its own 120-second controller. Therefore `Direct: aborted` is a separate direct-provider/timeout symptom; it should not be attributed to the parent-run signal without additional evidence. The confirmed regression here is the subprocess fallback cancellation coupling.

## Suggested lifecycle fix

The proposed `signal: undefined` change correctly decouples the turn signal, but removing the signal entirely also removes session ownership. A review could survive `/reload`, fork/resume, or shutdown and then write through stale stores or notify through a stale context.

I suggest:

- create one dedicated `AbortController` for background-review work for the lifetime of the session;
- pass its signal to both the direct review (`RunDirectMemoryCompletionOptions.signal`) and subprocess fallback;
- add an idempotent `session_shutdown` handler that aborts the controller and boundedly awaits the active review;
- guard memory mutations and UI notifications after session cancellation/stale context;
- retain the parent-turn-abort regression test and add a shutdown-cancellation test proving that the child is cancelled and no stale write/notification occurs.

This keeps background review cancellable at the session boundary without allowing an unrelated interactive-turn abort to kill it. The separate direct timeout and any genuine watchdog timeout should be tracked independently if they remain after this lifecycle fix.

xz-dev:
Implemented the requested session lifecycle ownership and rebased onto current `main`.

The background review now owns one session-scoped `AbortController`; its signal is passed to both direct and subprocess transports, never the interactive turn signal. An idempotent `session_shutdown` handler aborts once and boundedly awaits the active review. Cancelled sessions do not start fallback work or emit stale success/failure notifications.

The direct path now handles already-aborted signals, removes its abort listener in `finally`, and refuses parse/apply after cancellation even when a provider ignores abort. Cancellation is also checked after acquiring the Markdown mutation lock, preventing a queued old-session mutation from starting after shutdown. Once a write has actually started, it is not rolled back.

Shutdown waits up to a local one-second cleanup grace; ordinary review/provider execution retains the existing 120-second timeout. The separate direct-provider timeout symptom and expected `router/*` / `--no-extensions` warning remain outside this PR.

Validation on `ea56ed3`:
- focused lifecycle/direct/MemoryStore tests passed
- full suite: all 50 test files
- type check
- minimum SDK 0.80.6 check
- `git diff --check`

The added regressions fail on the previous revision, including turn-vs-session signal ownership, shutdown cancellation/silence, no fallback after shutdown, ignored-provider abort handling, bounded shutdown, and cancellation while waiting for the Markdown mutation lock.

xz-dev:
Follow-up for the failed Tests job on `ea56ed3`: the shutdown grace timer was incorrectly `unref()`ed. When the mocked review ignored abort and no unrelated handles remained, Node could end the event loop before the grace deadline, leaving the awaited shutdown promise pending. Local runs happened to retain incidental handles; CI exposed the lifecycle bug.

`7948444` removes `timer.unref()`, so the bounded deadline itself remains capable of completing shutdown.

Revalidation:
- isolated bounded-shutdown regression: 20/20 fresh processes passed
- `tests/handlers/background-review.test.ts`: 34/34
- focused lifecycle/direct/MemoryStore suite: 149/149
- full `npm test`: all 50 test files passed
- `npm run check`
- `npm run check:min-sdk` (SDK 0.80.6)
- clean-head and `git diff --check` gates

Pushed with an exact lease against `ea56ed33822bb23a3d31ee6fd3b3c10a41f19933`.

chandra447 (commented):
Adversarial review — 3/4 reviewers clean, 1 point to confirm with the author (non-blocking):

**Direct-timeout no longer falls back to subprocess?** (src/handlers/background-review.ts:297-302). The new `fallbackReason === 'aborted'` early-return sits under the `sessionCancelled()` check, so a *mid-session* 120s direct timeout now skips the subprocess fallback that main still runs. If intentional (avoid 240s of review work after a timeout), please state it — the PR intent says fallback is suppressed 'after session cancellation', which reads narrower than the code. If unintentional, scope the skip to `sessionCancelled()` (already checked directly above). Otherwise the session-scoping, 1s shutdown grace, mutation-boundary recheck, and notification suppression all look correct.

PR #200, opened 2026-08-24
fix: skip reload flush and add lifecycle timings

### LINKED ISSUE #199: /reload awaits a memory flush even though the session continues
## What happened

On my machine, `/reload` sometimes takes 5 to 15 seconds. I traced one blocking path: after six user messages, `pi-hermes-memory` awaits an LLM flush even though reload preserves the session.

Pi sends `reason: "reload"` during shutdown. The extension ignores the reason and runs the same flush used for a real shutdown. That flush may call the configured model and then fall back to a subprocess.

## Environment

- macOS 26.5.2, build 25F84
- Darwin 25.5.0, Apple M3, arm64
- Node.js 24.14.1
- Pi 0.84.2
- `pi-hermes-memory`: package version 0.9.6 at commit `71beae8`
- Main Hermes SQLite file: 212.3 MiB
- Pi session storage: 8.55 GiB
- Session archive: 4,115 JSONL files
- Files visible to the current non-recursive indexer: 1,506

My config changes only model and thinking settings. The shutdown-flush defaults are unchanged:

```ts
reviewTransport: "direct",
flushOnShutdown: true,
flushMinTurns: 6,
flushRecentMessages: 0,
```

## How to reproduce

1. Start Pi with the default shutdown-flush settings.
2. Send at least six user messages.
3. Run `/reload`.

I also reproduced the handler path with the real `setupSessionFlush()`. After six user `message_end` events, this shutdown event called the direct-completion dependency once:

```ts
{ type: "session_shutdown", reason: "reload" }
```

A provider can still stop before an outbound request if no model or authentication is available. The important part is that reload enters and awaits the flush path.

## Why reload waits

Pi awaits the shutdown event before rebuilding extensions:

```js
await emitSessionShutdownEvent(this._extensionRunner, {
  type: "session_shutdown",
  reason: "reload",
});
```

The extension currently ignores the reason:

```ts
pi.on("session_shutdown", async (_event, ctx) => {
  if (!config.flushOnShutdown) return;
  await flush(ctx, undefined, 10000);
});
```

The direct and subprocess attempts have separate time limits. If direct completion fails, the subprocess gets a new timeout. Reload waits until the handler finishes.

## Other work in the same lifecycle

The flush is not the only synchronous work around startup and shutdown. The extension also:

- reconciles Markdown memory files and loads stores;
- scans session metadata and may parse full JSONL files;
- rereads the active session during live and shutdown indexing;
- runs `PRAGMA wal_checkpoint(TRUNCATE)` when closing the database;
- schedules `PRAGMA quick_check` after opening the database.

`setTimeout(0)` delays some of this work, but the callback still runs on the UI thread.

I found no current backfill backlog. All 1,506 index-visible files had matching size and modification-time metadata. A read-only `quick_check` on this database took 2.25 seconds during the initial check.

I also measured these startup times:

| Configuration | Ready time |
|---|---:|
| Full Pi, warm runs | 3.5 to 4.0s |
| Full Pi, cold outlier | 12.2s |
| No extensions, online | 0.74s |
| No extensions, offline | 0.56s |

With `PI_TIMING=1`, extension imports took 1.454s in total. `pi-hermes-memory` took about 481ms.

## Suggested change

Skip the shutdown flush only for reload:

```ts
pi.on("session_shutdown", async (event, ctx) => {
  if (!config.flushOnShutdown || event.reason === "reload") return;
  await flush(ctx, undefined, 10000);
});
```

Keep direct completion and subprocess fallback for real shutdowns.

It would also help to add opt-in `PI_TIMING=1` measurements around startup, indexing, flush, checkpoint, integrity scan, and database close. Those numbers can guide any later indexing or SQLite changes.

## Checks

1. Reload calls neither direct completion nor the subprocess fallback.
2. Other shutdown reasons keep the current flush behavior.
3. Fewer than `flushMinTurns` user messages still skips the flush.
4. Timing stays silent unless `PI_TIMING=1` is set.
5. Timing does not change return values or errors.
6. The type-check and test suite passes.

## Related work

- [Issue #83](https://github.com/chandra447/pi-hermes-memory/issues/83) and [PR #84](https://github.com/chandra447/pi-hermes-memory/pull/84) bounded startup backfill.
- [PR #124](https://github.com/chandra447/pi-hermes-memory/pull/124) reduced SQLite connection churn during `/new` and `/resume`.
- [PR #186](https://github.com/chandra447/pi-hermes-memory/pull/186) moved full-database integrity scans out of synchronous open.
- Commit [`58c9cef`](https://github.com/chandra447/pi-hermes-memory/commit/58c9cef13e41b161c249ab0a45910c7e37c18da8) made shutdown flush fire-and-forget. Commit [`77b6430`](https://github.com/chandra447/pi-hermes-memory/commit/77b64305a17d06e296eceef41abc2a786b55428e) restored the awaited handler without filtering reload.

## Workaround

```json
{
  "flushOnShutdown": false
}
```

This disables the flush for every shutdown reason, not only reload.

AndriiSemenkov:
## What changed

- Skip the shutdown memory flush when Pi reports `reason: "reload"`.
- Keep direct completion and subprocess fallback unchanged for other shutdown reasons.
- Add opt-in `PI_TIMING=1` measurements for startup, indexing, shutdown, and SQLite work.
- Document the timing labels and reload behavior.

## Verification

- `npm run check:min-sdk && npm run check && npm test`
- All 48 test files passed.
- A live Pi 0.84.2 smoke test reloaded the local extension without a `shutdown.flush` span and exited with code 0.

Fixes #199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Opt-in memory lifecycle timings**: `PI_TIMING=1` now reports startup synchronization and loading, backfill and live-index callbacks, shutdown flush/index/wait/close, and database open/integrity-check/checkpoint durations. Normal runs produce no timing output.
+
 ### Fixed
 
+- **Reload no longer waits for an unnecessary LLM memory flush**: Pi preserves the current session during `/reload`, but the shutdown handler ignored `event.reason` and awaited direct completion plus possible subprocess fallback after six user turns. Reload now skips both transports; quit and session-replacement reasons keep the existing flush policy.
 - **Portable child extension sources were discarded before Pi could resolve them**: `childExtensionPaths` ran every configured value through Node's cwd-relative `path.resolve()` and required that resulting local path to already exist. This silently dropped Pi-native `~/...`, `git:...`, and `npm:...` extension sources, forcing machine-specific absolute paths for custom providers in isolated subprocesses. Configured sources are now passed unchanged to Pi's standard `-e` resolver, while Hermes-discovered auth adapter paths retain their local existence checks.
 
 ## [0.9.4] - 2026-08-08
--- a/README.md
+++ b/README.md
@@ -558,6 +558,24 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `flushMinTurns` | `6` | Minimum turns before flush triggers |
 | `flushRecentMessages` | `0` | Recent messages included in session flush (`0` = all) |
 
+## Diagnosing lifecycle latency
+
+Run Pi with timing enabled to see which memory lifecycle step is slow:
+
+```bash
+PI_TIMING=1 pi
+```
+
+`pi-hermes-memory` writes these spans to stderr only when timing is enabled:
+
+- `session-start.persistence-sync` and `session-start.load`
+- `session-backfill.check` and `session-backfill.callback`
+- `live-index.callback`
+- `shutdown.flush`, `shutdown.active-index`, `shutdown.index-waits`, and `shutdown.database-close`
+- `database.open`, `database.quick-check`, and `database.checkpoint`
+
+The deferred backfill, live-index, and integrity-check spans may appear after startup spans because they run on later timer turns. `/reload` does not run `shutdown.flush`; other shutdown reasons keep the configured direct completion and subprocess fallback. Use the measured spans before changing indexing, checkpoint, or synchronization policy.
+
 ## Where Data Lives
 
 ```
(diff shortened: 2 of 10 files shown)


chandra447:
Thanks for the focused reload fix and the lifecycle timing instrumentation. The change follows Pi's documented `session_shutdown` reason contract, skips the awaited LLM flush only during reload, and keeps timing output opt-in through `PI_TIMING=1`. The current head is clean and all repository checks pass. I am completing the final independent verification before merge.

chandra447:
Independent verification passed at head `212ee3415b9957db14971ebbc56be37065634adc`. The focused PR tests passed, and a real Pi CLI smoke exercised `/reload` and exited with code 0. Thanks again for the careful fix. This PR is now merged as `8720c571c4946aea0faacb23fd286a6d193bbf91`.

PR #207, opened 2026-08-28
fix: preserve parent context in background reviews

XiaYiHann:
## Summary

- Reuse Pi's `buildSessionContext()`/`convertToLlm()` for eligible full-history direct background reviews.
- Keep subprocess fallback branch-, compaction-, and custom-message-aware instead of flattening raw session entries.
- Inherit the active thinking level only for the direct/fallback path; forced subprocess transport does not impose an implicit `thinking: off`.
- Add configurable `quickCheckOnOpen` to avoid synchronous SQLite startup stalls while retaining operation-time recovery.
- Add regression tests and document the transport/thinking behavior.

## Validation

- `TMPDIR=/var/tmp npm test` — all 49 test files passed
- `TMPDIR=/var/tmp npm run check`
- `TMPDIR=/var/tmp npm run check:min-sdk` — passes against `@earendil-works/pi-coding-agent@0.80.1`
- `git diff --check`


chandra447 (changes requested):
Thank you for using pi-hermes-memory and for contributing the parent-context preservation, transport parity, thinking-level, and quick-check configuration work in this PR. Reusing Pi's session-context helpers is the right direction, and I verified that the touched Pi APIs are public and available at the declared minimum SDK. I am requesting changes before merge for these behavior issues.

1. Empty direct responses now skip the subprocess fallback.

`runDirectMemoryCompletion()` returns `{ ok: true, fallbackReason: "empty" }` when the provider returns no final text at `src/handlers/review-memory-ops.ts:499-502`. `setupBackgroundReview()` returns immediately on `ok` at `src/handlers/background-review.ts:361-368`, so the review is silently dropped. Before this change, empty text became a parse failure and reached the subprocess fallback. Keep an explicit parsed `{"operations":[]}` as the deliberate no-op, but treat missing final content as a failure or a distinct `no_content` result that falls back.

2. Thinking blocks are parsed as executable memory operations.

`responseText()` falls back to concatenated thinking blocks at `src/handlers/review-memory-ops.ts:404-421`, and the result goes through `parseReviewOperations()`. Thinking is the model's scratch reasoning, not its final answer. A draft or hypothetical JSON object in that channel can now mutate MEMORY.md, USER.md, or failure memory. Please accept operations only from final text, or require a strict whole-response JSON object before accepting any thinking-only response.

3. Removing the implicit `thinking: off` default changes every child transport.

The shared `execChildPrompt()` helper is used by background review, auto-consolidation, correction saves, and session flushes. Changing `effectiveThinkingOverride()` at `src/handlers/pi-child-process.ts:86-88` means a configured `llmModelOverride` without `llmThinkingOverride` now uses the provider default for all of those calls, not only background review fallback. That changes cost and output behavior for consolidation and correction, while the README and changelog describe only the review path. Either scope the change to review fallback or document and test the wider behavior explicitly.

4. Disabling open-time quick_check leaves common reads without recovery.

With `quickCheckOnOpen: false`, `memory_search`, `session_search`, and their statistics reads call `getDb()` and search directly. They do not use `withCorruptionRecovery()`. A corrupt page can therefore produce repeated raw search errors instead of being quarantined and rebuilt, even though the README says operation-time recovery remains enabled. Wrap the public read boundary or narrow the documentation to the operations that actually recover.

5. The option is not applied to the manual session-index command.

`src/handlers/index-sessions.ts:38` constructs a separate `DatabaseManager` with default options. A user who sets `quickCheckOnOpen: false` still gets the open-time scan when running `/memory-index-sessions`. Pass the configured option through or reuse the main manager.

6. The subprocess fallback still truncates the SDK-built context.

`collectLlmMessageParts()` calls `getMessageText()` with its default 500-character limit. Long compaction and branch summaries are therefore truncated in the fallback path, despite the README saying the SDK-built context is preserved. Please use an intentional limit for this path and add a regression covering a summary longer than 500 characters.

I checked the Pi extension contracts against the current docs and SDK. `turn_end`, `ctx.sessionManager.getBranch()`, `getLeafId()`, `ctx.getSystemPrompt()`, `buildSessionContext()`, `convertToLlm()`, and `pi.exec()` are valid public constructs, and the PR passes the minimum SDK check. GitHub CI also passes Type Check, Min Supported SDK, Tests, and Lint. The requested changes are about result handling, transport semantics, and recovery coverage.

Issue #201, opened 2026-08-25
v0.9.6 typecheck fails with Pi SDK 0.84.3 due to ResolvedRequestAuth drift


oversk7:
## Summary

`pi-hermes-memory@0.9.6` declares Pi SDK peer dependencies as `>=0.80.1`, but its typecheck fails when the development SDK packages are updated from `0.80.2` to `0.84.3`.

This is not caused by Hermes configuration or memory data. The local `ResolvedRequestAuth` mirror in `review-memory-ops.ts` has drifted from the SDK type: current `ProviderHeaders` allows `string | null`, while the local type only allows `string`.

## Reproduction

```sh
git clone --depth 1 --branch v0.9.6 https://github.com/chandra447/pi-hermes-memory.git
cd pi-hermes-memory
npm ci
npm install --no-save --package-lock=false \
  @earendil-works/pi-ai@0.84.3 \
  @earendil-works/pi-coding-agent@0.84.3 \
  @earendil-works/pi-tui@0.84.3
npm run check
```

## Actual result

```text
src/handlers/review-memory-ops.ts(164,3): error TS2322:
Type '...ResolvedRequestAuth' is not assignable to type '...ResolvedRequestAuth'.
  Types of property 'headers' are incompatible.
    Type 'ProviderHeaders | undefined' is not assignable to type
    'Record<string, string> | undefined'.
      Type 'string | null' is not assignable to type 'string'.
        Type 'null' is not assignable to type 'string'.
```

`npm run check` exits with status 2.

## Expected result

The project should typecheck against SDK versions accepted by its peer dependency range, including the current `0.84.3` release.

## Baseline

Reinstalling the repository's pinned development SDK versions (`0.80.2`) makes `npm run check` pass.

## Suggested direction

- Use the SDK's `ProviderHeaders` type instead of `Record<string, string>` in the local auth type mirror.
- Review the rest of the mirrored `ResolvedRequestAuth` shape for drift (the current SDK also includes `baseUrl`).
- Add CI coverage for both the minimum supported SDK and the current/latest SDK so later 0.x type changes are caught.

## Environment

- pi-hermes-memory: `0.9.6` / tag `v0.9.6`
- Pi SDK packages: `0.84.3`
- Node.js: `24.14.1`
- OS: Windows

chandra447:
[...]

PR #158, opened 2026-08-08
feat(config): accept "max" as llmThinkingOverride

ustc21xyx:
## Summary

`@earendil-works/pi-ai` added `"max"` to `ThinkingLevel` in **0.81.0**, but this extension keeps its own local `ThinkingLevel` union and `THINKING_LEVELS` validation list, both stopping at `"xhigh"`. As a result, `llmThinkingOverride: "max"` in `hermes-memory-config.json` is silently ignored by `isThinkingLevel()` and the override falls back to the default (`"off"` when `llmModelOverride` is set) — with no warning surfacing the real cause.

This affects providers/models whose top reasoning tier is exposed as `max`.

## Changes

- `src/types.ts`: add `"max"` to the `ThinkingLevel` union
- `src/config.ts`: add `"max"` to `THINKING_LEVELS` so config validation accepts it
- `tests/config.test.ts`: include `"max"` in the valid `llmThinkingOverride` loop
- `tests/handlers/review-memory-ops.test.ts`: prove `llmThinkingOverride: "max"` reaches the direct `completeSimple` reasoning option
- `tests/handlers/pi-child-process.test.ts`: prove `llmThinkingOverride: "max"` becomes child `--thinking max`
- `README.md`: document `max` in the `llmThinkingOverride` row
- `package.json`: raise `@earendil-works/pi-ai` and `@earendil-works/pi-coding-agent` peers to `>=0.81.0`, and align development SDKs (`pi-ai`, `pi-coding-agent`, `pi-tui`) to `^0.81.0`. `pi-tui` is not a peer, but its types cross `ExtensionCommandContext.ui.custom`; leaving it on 0.80.x produces a duplicate-private-property error against 0.81.

Auth stays on main's public `modelRegistry.getApiKeyAndHeaders()` path. This branch does not restore structural access to private `authStorage`.

## Verification

- `npm run check` (tsc --noEmit): clean
- `npm run check:min-sdk`: type-checks against `pi-coding-agent@0.81.0`
- `npm test`: all 47 test files passed
- lockfile regenerated with npm 10 (CI toolchain)

PR #209, opened 2026-08-29
fix: derive thinking/auth types from Pi SDK

### LINKED ISSUE #198: Support `max` in `llmThinkingOverride`
## Summary

`llmThinkingOverride` currently accepts values only up to `xhigh`, while current Pi versions also support the `max` thinking level.

More importantly, configuring `"llmThinkingOverride": "max"` is silently ignored. When `llmModelOverride` is also set, the maintenance calls then fall back to `thinking=off`, which is the opposite of the user's intent.

## Reproduction

With pi-hermes-memory 0.9.6, configure:

```json
{
  "llmModelOverride": "openai-codex/gpt-5.6-luna",
  "llmThinkingOverride": "max"
}
```

Reload Pi and let a background review, correction save, session flush, or consolidation run.

## Actual behavior

The local type and validator exclude `max`:

- `src/types.ts`: `ThinkingLevel` ends at `xhigh`
- `src/config.ts`: `THINKING_LEVELS` ends at `xhigh`

As a result, `max` is discarded during config parsing. Because a model override remains configured without a valid thinking override, `effectiveThinkingOverride()` in the direct/subprocess paths defaults to `off`.

This failure is silent, so users may believe the background memory task is running at `max` while it is actually running at `off`.

## Expected behavior

`max` should be accepted and forwarded to both direct and subprocess maintenance calls, consistent with Pi's current `ThinkingLevel` type:

```ts
"off" | "minimal" | "low" | "medium" | "high" | "xhigh" | "max"
```

## Suggested change

- Add `max` to the local `ThinkingLevel` union in `src/types.ts`
- Add `max` to `THINKING_LEVELS` in `src/config.ts`
- Update the README configuration table
- Add a config parsing test that verifies `max` is preserved and forwarded
- Optionally warn on unknown thinking levels rather than silently dropping them

## Environment

- pi-hermes-memory: 0.9.6
- Pi: 0.84.2
- Model used for the attempted override: `openai-codex/gpt-5.6-luna`

chandra447:
This is the remaining review notes from [#158](https://github.com/chandra447/pi-hermes-memory/pull/158) (@ustc21xyx), not a new feature.

Fixes [#198](https://github.com/chandra447/pi-hermes-memory/issues/198) and [#201](https://github.com/chandra447/pi-hermes-memory/issues/201).

## What changed

- Stop mirroring Pi SDK unions. `ThinkingLevel` is now `ModelThinkingLevel` from `@earendil-works/pi-ai`. JSON parsing keeps a local runtime list so `"max"` is accepted even when typechecking against the locked 0.80.2 SDK (which lacks `max`; Pi added it in **0.80.6**).
- `ResolvedRequestAuth` is derived from `modelRegistry.getApiKeyAndHeaders()` instead of a local `Record<string, string>` headers mirror, so `ProviderHeaders` (`string | null`) and `baseUrl` track the installed SDK. Auth stays on that public registry API; private `authStorage` is not used. `baseUrl` is not passed into `completeSimple` (StreamOptions has no such field).
- Peer floor for `@earendil-works/pi-ai` and `@earendil-works/pi-coding-agent` is `>=0.80.6`. Not raised to 0.81.0. DevDependencies / `pi-tui` unchanged.
- Tests: config parsing accepts `"max"`; `llmThinkingOverride: "max"` is asserted on direct `completeSimple` reasoning **and** child `--thinking max`.

## CI

Please prove **check**, **tests**, and **min-sdk**.

**Lockfile follow-up:** GitHub MCP `push_files` cannot round-trip `package-lock.json` (144k; tool results truncate ~40k). The only intended lockfile edit is `packages[""].peerDependencies` `>=0.80.1` → `>=0.80.6` (two strings). The lint job (`npm install` then `git diff --exit-code`) will fail until that snapshot is updated. Do not regenerate the rest of the lockfile.

README table still lists thinking levels through `xhigh`; docs can follow once the lockfile lands.


ustc21xyx (approved):
Reviewed the full diff and relevant SDK declarations. The implementation looks correct:

- `ThinkingLevel` now tracks `ModelThinkingLevel`; the runtime parser explicitly accepts `max`.
- Verified `max` is absent in Pi SDK 0.80.2 and present in 0.80.6, so the peer floor is accurate.
- Auth result typing is derived from the public `modelRegistry.getApiKeyAndHeaders()` API and preserves SDK header/env compatibility.
- Direct completion and child-process propagation of `max` are both covered.

Local verification on this head:
- `npm run check` — passed
- `npm run check:min-sdk` — passed against 0.80.6

Non-code follow-up before merge: please commit the two `package-lock.json` root peerDependency snapshot changes (`>=0.80.1` → `>=0.80.6`) so the Lint job turns green. Running `npm install` produced exactly those two replacements and no dependency-tree churn.

PR #213, opened 2026-08-30
fix(search): accept target:"project" filter in memory_search

### LINKED ISSUE #212: memory_search schema rejects target:"project" filter though write path and result labels use it
## Summary

`memory_add` accepts `target: "project"` for writes, and `memory_search` labels project-scoped results `[target=project]` (the mutation target that `memory_replace` / `memory_remove` require). The `memory_search` tool schema, however, only declares `['memory', 'user', 'failure']` for its `target` **filter**, so querying by the very target value search results display fails schema validation before the handler runs.

## Reproduction (v0.9.7)

1. Store a project memory: `memory_add(scope: "project", target: "memory", content: "deploy uses pnpm")`.
2. Ask an agent to search only project-attributed entries: `memory_search(query: "pnpm", target: "project")`.
3. The harness validates arguments against the tool schema and rejects the call:

```text
Validation failed for tool "memory_search":
  - target: must be equal to one of the allowed values
```

4. The workaround is non-obvious: the model must know to pass `project: "<name>"` instead, and that works only when the project name is known in advance — there is no filter for "any project".

## Root cause

Three surfaces disagree on the meaning of `target`:

| Surface | Values |
|---|---|
| `memory_add` write target (guidance in `constants.ts`) | `memory`, `user`, `project`, `failure` |
| `memory_search` result label (`mutationTarget()`: `memory` + `project` → `project`) | includes `project` |
| `memory_search` filter schema (`memory-search-tool.ts`) | `memory`, `user`, `failure` — no `project` |

Project-attributed memory entries are stored as `target='memory'` with `project != NULL`, so there is no `target='project'` row; the filter needs a mapping, not just a schema widening.

## Expected behavior

- `memory_search` accepts `target: "project"`.
- It resolves to the entries whose displayed mutation target is `project`, i.e. `target = 'memory' AND project IS NOT NULL` — project-attributed failures (`[target=failure]`) stay in the `failure` store and are not matched, keeping filters consistent with what `memory_replace` / `memory_remove` consume.
- The injected memory-policy guidance lists the fourth accepted value.

## Regression tests

1. Store-level: seeded global + project entries; `searchMemories(query, { target: 'project' })` returns only project-attributed rows; combining with `project: '<name>'` narrows; `project: null` returns nothing (those rows are project-scoped).
2. Tool-level: register the tool with a mock `pi`, assert `parameters.properties.target.enum` contains `'project'`, and `execute()` with `target: 'project'` returns only `[target=project]` results.

manooog:
Closes #212

## Problem

`memory_add` accepts `target: "project"` for writes, and `memory_search` labels project-scoped results `[target=project]` — the exact value `memory_replace` / `memory_remove` then require. But the `memory_search` **filter** schema only declared `['memory', 'user', 'failure']`, so querying by the mutation target that the tool itself displays failed harness-side schema validation before the handler ran:

```text
Validation failed for tool "memory_search":
  - target: must be equal to one of the allowed values
```

Project-attributed memory entries are stored as `target='memory'` with `project != NULL` — there is no `target='project'` row — so this needed a filter mapping, not just a schema widening.

## Changes

- **`src/tools/memory-search-tool.ts`** — `target` filter enum gains `'project'`; execute args typed to the four-value union; tool description documents the filter.
- **`src/store/sqlite-memory-store.ts`** — `buildScopeConditions()` and the inline `searchMemories()` conditions map `target === 'project'` to `target = 'memory' AND project IS NOT NULL`. This matches `mutationTarget()` exactly: the filter returns precisely the entries whose displayed mutation target is `project`. Project-attributed failures stay `[target=failure]` (they live in the failure store and are mutated through `target: "failure"`) and are intentionally not matched. Composable with the existing `project` param (`'<name>'` narrows; `null` returns nothing).
- **`src/constants.ts`** — injected memory-policy guidance lists the fourth accepted filter value and mentions `target="project"` for narrowing when the project name is unknown.
- **`tests/store/sqlite-memory-store.test.ts`** — store-level regression: project-only results for `target: 'project'`, unchanged behavior for `target: 'memory'`, and composition with `project: '<name>'` / `project: null`.
- **`tests/tools/memory-search-tool.test.ts`** — tool-level regression: schema enum contains `'project'`; `execute()` with `target: 'project'` returns only `[target=project]` results.
- **`CHANGELOG.md`** — entry under `[Unreleased] → Fixed`.

## Verification

- `npm run check` (tsc, Node 24 — CI matrix is 22) ✅
- `bash tests/run-all.sh` — all 50 test files pass (735 tests, 3 of them new store-level, 1 new tool-level) ✅

Schema shape that previously rejected the call, verified in tests: `parameters.properties.target.enum` now `['memory', 'user', 'failure', 'project']`.


src/store/sqlite-memory-store.ts
```diff
@@ -727,7 +730,10 @@ export function searchMemories(
       }
     }
 
-    if (target) {
+    if (target === 'project') {
+      conditions.push("m.target = 'memory'");
+      conditions.push('m.project IS NOT NULL');
+    } else if (target) {
```
chandra447: FTS mapping matches `mutationTarget()`: `target: "project"` means `m.target = 'memory' AND m.project IS NOT NULL`, and project-attributed failures stay `[target=failure]`. Good.

`runShortCjkFallback` a few lines below still does `m.target = ?` with the raw `'project'` string. A 1–2 character CJK query with `target: "project"` will miss after FTS returns empty. Apply the same helper there, and add a store-level test with a short CJK query so this path cannot drift again.

src/store/sqlite-memory-store.ts
```diff
@@ -131,7 +131,10 @@ function mapRow(row: {
 function buildScopeConditions(params: unknown[], target?: string, project?: string | null, category?: MemoryCategory | null): string[] {
   const conditions: string[] = [];
 
-  if (target) {
+  if (target === 'project') {
+    conditions.push("target = 'memory'");
+    conditions.push('project IS NOT NULL');
+  } else if (target) {
```
chandra447: This mapping belongs on the search paths only. `buildScopeConditions` is the helper for `syncMemoryEntry` / `replaceSyncedMemories` / `removeSyncedMemories` / `reconcileMarkdownMemoryScope`, whose `target` is `'memory' | 'user' | 'failure'`. Mutations already map `project` to SQLite `memory` plus a project column before they get here.

Leave this helper as a storage-column filter. Extract one search helper and use it from both `runSearch` and `runShortCjkFallback`.

chandra447 (changes requested):
Thanks for this. The schema widening and the FTS mapping match `mutationTarget()`: `target: "project"` should mean the displayed mutation target (`target = 'memory' AND project IS NOT NULL`), and project-attributed failures stay `[target=failure]`. Tests cover the enum, the label, `target: "memory"` still returning globals, and composition with `project: null`.

Two changes requested:

1. `runShortCjkFallback` still binds `m.target = ?` with the raw `'project'` string. A 1–2 character CJK query with `target: "project"` will miss after FTS returns empty. Use the same mapping there. A store-level test with a short CJK query would lock it.

2. Do not put this mapping in `buildScopeConditions`. That helper is for sync/replace/remove/reconcile, whose targets are `'memory' | 'user' | 'failure'` only. Extract one helper and use it from both FTS `runSearch` and the CJK LIKE path. Leave `buildScopeConditions` as a storage-column filter.

CI is green on `17f9501`. Happy to take this once the mapping is one place and covers both search paths.

src/store/sqlite-memory-store.ts
manooog: Fixed in 42a82f9: extracted `buildSearchTargetConditions` and wired it into `runShortCjkFallback`. Added store test `applies target "project" in the CJK literal fallback path` (2-char query `项目`).

src/store/sqlite-memory-store.ts
manooog: Fixed in 42a82f9: reverted `buildScopeConditions` to storage-column filtering only; `target: "project"` mapping lives in `buildSearchTargetConditions`, used by both `runSearch` and `runShortCjkFallback`.

chandra447 (approved):
Both requested changes landed in `42a82f9`.

`buildSearchTargetConditions` is the one mapping, used by FTS `runSearch` and `runShortCjkFallback`. `buildScopeConditions` is back to a storage-column filter for sync/replace/remove. The new store test (`项目`, 2 chars) covers the CJK LIKE path. Schema, labels, and `target: "memory"` behavior are unchanged from the first commit.

CI is green. Leftover: the PR body still says `buildScopeConditions` does the mapping. Not a code issue.

Approve, squash-merge.

chandra447:
Thanks for using the extension, and for taking the time to craft this. Grateful for people who notice a real gap, write it up, and ship a careful contribution like this.

PR #215, opened 2026-08-31
feat: llmFallbackModels chain for memory reviews

dominic-codespoti:
## Summary
Memory reviews (`background review`, `flush`, `consolidation`, `correction`) currently use a single `llmModelOverride`. If that model is disabled/401/429/500, the review fails even though other providers are available. This mirrors `Hermes` `fallback_providers` semantics for auxiliary tasks.

This PR adds a **model chain** for memory LLM calls.

## Changes
- **types.ts:** `MemoryConfig.llmFallbackModels?: string[]` — fallback chain tried in order after primary
- **config.ts:** parse `llmFallbackModels` + aliases `llmModelFallbacks` / `fallbackModels`; also supports array form `llmModelOverride: ["primary/fallback1", "fallback2"]`
- **review-memory-ops.ts:**
  - `resolveReviewModels()` / `getReviewModelChain()` — resolve `["provider/model"]` refs via `modelRegistry.getAll()`
  - `runDirectMemoryCompletion()` now iterates `primary + fallbacks` with per-model timeout/auth, falling through on `no_auth`/`provider_error`/`aborted`/`parse_error` to next model

## Example config
```json
{
  "llmModelOverride": "opencode-go/muse-spark-1.2-contributor",
  "llmFallbackModels": [
    "opencode-go/deepseek-v4-flash",
    "openrouter/anthropic/claude-sonnet-4"
  ]
}
```
or
```json
{ "llmModelOverride": ["opencode-go/muse-spark", "opencode-go/deepseek-v4-flash"] }
```

## Compatibility
- Fully backward compatible; no config = existing single-model behavior
- Falls back to active `ctx.model` if chain doesn't resolve
- Mirrors `hermes fallback_providers` per-turn semantics (chain is per-review, not persisted)

Fixes silent memory review failure when single override is disabled.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -157,6 +157,28 @@ export function loadConfig(configPath = DEFAULT_CONFIG_PATH): MemoryConfig {
         const trimmed = parsed.llmModelOverride.trim();
         if (trimmed.length > 0) config.llmModelOverride = trimmed;
       }
+      // Support array form for primary override too (e.g. llmModelOverride: ["a/b","c/d"]) — first entry is primary, rest are fallbacks
+      if (Array.isArray(parsed.llmModelOverride) && parsed.llmModelOverride.every((v: unknown) => typeof v === "string")) {
+        const cleaned = (parsed.llmModelOverride as string[]).map((s) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) {
+          config.llmModelOverride = cleaned[0];
+          const fallbacks = cleaned.slice(1);
+          if (fallbacks.length > 0) config.llmFallbackModels = fallbacks;
+        }
+      }
+      if (isStringArray(parsed.llmFallbackModels)) {
+        const cleaned = (parsed.llmFallbackModels as string[]).map((s) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) config.llmFallbackModels = [...new Set(cleaned)];
+      }
+      // Backward-compat alias: llmModelFallbacks / fallbackModels
+      if (isStringArray((parsed as Record<string, unknown>).llmModelFallbacks)) {
+        const cleaned = ((parsed as Record<string, unknown>).llmModelFallbacks as string[]).map((s: string) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) config.llmFallbackModels = [...new Set([...(config.llmFallbackModels ?? []), ...cleaned])];
+      }
+      if (isStringArray((parsed as Record<string, unknown>).fallbackModels)) {
+        const cleaned = ((parsed as Record<string, unknown>).fallbackModels as string[]).map((s: string) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) config.llmFallbackModels = [...new Set([...(config.llmFallbackModels ?? []), ...cleaned])];
+      }
       if (isThinkingLevel(parsed.llmThinkingOverride)) config.llmThinkingOverride = parsed.llmThinkingOverride;
       if (isStringArray(parsed.childExtensionPaths)) {
         const childExtensionPaths = [...new Set<string>(
--- a/src/handlers/review-memory-ops.ts
+++ b/src/handlers/review-memory-ops.ts
@@ -48,7 +48,7 @@ export function usesDirectTransport(config: Pick<MemoryConfig, "reviewTransport"
   return (config.reviewTransport ?? "direct") === "direct";
 }
 
-type ReviewLlmConfig = Pick<MemoryConfig, "llmModelOverride" | "llmThinkingOverride">;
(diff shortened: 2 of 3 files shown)


src/handlers/review-memory-ops.ts
```diff
@@ -398,109 +436,126 @@ export async function runDirectMemoryCompletion(
...
+        );
+      }
 
-    if (response.stopReason === "aborted") {
-      return { ok: false, appliedCount: 0, fallbackReason: "aborted" };
-    }
+      if (response.stopReason === "aborted") {
+        lastResult = { ok: false, appliedCount: 0, fallbackReason: "aborted" };
+        if (mi < models.length - 1) { clearTimeout(timeout); continue; }
+        return lastResult;
```
chandra447: Do not continue the chain on `aborted`.

`options.signal` is still the interactive turn (see #192). Esc, session shutdown, or the 120s watchdog all abort this controller. Hopping to the next model after that fights the session lifecycle and can turn one timeout into N sequential 120s calls.

Return `aborted` immediately. Same in the `catch` below, which also `continue`s after `controller.signal.aborted`.

Retry only on `no_auth` and provider/transport failures. Add a test that an aborted primary does not invoke the fallback model.

src/config.ts
```diff
@@ -157,6 +157,28 @@ export function loadConfig(configPath = DEFAULT_CONFIG_PATH): MemoryConfig {
...
+      }
+      // Backward-compat alias: llmModelFallbacks / fallbackModels
+      if (isStringArray((parsed as Record<string, unknown>).llmModelFallbacks)) {
+        const cleaned = ((parsed as Record<string, unknown>).llmModelFallbacks as string[]).map((s: string) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) config.llmFallbackModels = [...new Set([...(config.llmFallbackModels ?? []), ...cleaned])];
+      }
+      if (isStringArray((parsed as Record<string, unknown>).fallbackModels)) {
+        const cleaned = ((parsed as Record<string, unknown>).fallbackModels as string[]).map((s: string) => s.trim()).filter(Boolean);
+        if (cleaned.length > 0) config.llmFallbackModels = [...new Set([...(config.llmFallbackModels ?? []), ...cleaned])];
+      }
```
chandra447, line 189: One config key. `llmFallbackModels` is enough.

This also accepts `llmModelOverride` as an array, plus `llmModelFallbacks`, plus `fallbackModels`, and they merge. There is no old published field to keep compatible with. Array-vs-string `llmModelOverride` also makes the data shape illegal states representable (a string primary vs a list that silently overwrites fallbacks depending on parse order).

Keep `llmModelOverride: string` and `llmFallbackModels: string[]`. Drop the aliases and the array form.

chandra447 (changes requested):
Thanks for this. The idea is in scope. Auth still goes through `modelRegistry.getApiKeyAndHeaders`, and CI is green on `70e434b`.

Not mergeable yet.

1. Do not continue the chain on `aborted`. Esc, session shutdown, and the per-model 120s watchdog all abort the same controller (`options.signal` is still the turn; see #192). Hopping to the next model after abort fights that lifecycle. Return `aborted` immediately, including in the `catch`. Retry only on `no_auth` and provider/transport failures.

2. One config key. Keep `llmModelOverride: string` and `llmFallbackModels: string[]`. Drop `llmModelFallbacks`, `fallbackModels`, and the array form of `llmModelOverride`. There is no old published alias to preserve.

3. Add tests. At least: abort on the primary does not call the fallback; `no_auth` on the primary does. The diff does not touch `tests/`.

Happy to take this once abort stops the chain and the config shape is one list.

chandra447 (commented):
Adversarial review (3/4 reviewers converged) — 2 blocking issues:

1. **Fallback chain ignores caller abort** (src/handlers/review-memory-ops.ts ~502-554). `stopReason: aborted` and `controller.signal.aborted` fall through to the next model, which builds a *fresh* AbortController and only `addEventListener('abort')` — never fires for an already-aborted signal. So a user cancel / 10s session-flush bound retries every remaining model for another full 120s each (and can still `applyReviewOperations`). Fix: check `options.signal.aborted` at the top of each iteration and return `aborted` instead of continuing; only fall through on per-model timeout/provider errors when the caller is still alive.
2. **Array-form fallbacks silently dropped when the canonical key is also set** (src/config.ts:160-172). `llmModelOverride: ['a/b','c/d']` seeds fallbacks with `[c/d]`, but a subsequent `llmFallbackModels: ['e/f']` *assigns* `[...new Set(cleaned)]`, discarding `c/d` (aliases merge, the canonical key doesn't). Fix: merge instead of assign.

Process note: this PR, #191, and #192 all rewrite `runDirectMemoryCompletion`'s auth/abort sections — expect conflicts; suggest landing #191 + #192 first, then rebasing this chain on top.

Issue #211, opened 2026-08-29
Auto-consolidation cannot resolve child CLI under standalone OMP on Windows


richard-scott:
## Summary

Automatic over-capacity consolidation fails when `pi-hermes-memory` runs inside standalone OMP on Windows because the subprocess resolver requires a directly executable Pi CLI. OMP provides `omp.exe`, not a separate `pi.exe` or resolvable Pi `cli.js`.

This appears related to #69, but the fix for that issue does not cover a standalone OMP host.

## Environment

- Windows 11 Pro, build 26200
- OMP 18.0.11 (`%LOCALAPPDATA%\omp\omp.exe`)
- `pi-hermes-memory` 0.9.6
- Node 24.14.0
- No `pi` executable on `PATH`; `omp.exe` is available

## Observed behavior

When the `failure` memory store reaches capacity and triggers automatic consolidation, the harness reports:

```text
⚠️ Auto-consolidation failed for 'failure': Consolidation failed: Error: Unable to resolve a directly executable Pi CLI on Windows
```

The memory write that triggered over-capacity consolidation cannot complete. The interactive OMP session otherwise continues.

## Reproduction

1. Run OMP 18.0.11 on Windows with `pi-hermes-memory` 0.9.6 enabled.
2. Do not install a separate Pi CLI; only the standalone `omp.exe` host is present.
3. Add entries until the `failure` memory store reaches its character limit and automatic consolidation runs.
4. Observe the warning above.

## Root cause

The automatic consolidator registered in `src/index.ts` calls `triggerConsolidation` without `directCtx`. As documented in `src/handlers/auto-consolidate.ts`, the automatic over-capacity path is therefore subprocess-only.

`resolveChildPiInvocation` in `src/handlers/pi-child-process.ts` then looks for a Pi `cli.js`, `pi.exe`/`pi.com`, or a `pi.cmd`/`pi.bat` accompanied by the Pi package's `cli.js`. None exists in a standalone OMP installation, so it throws.

## Expected behavior

Automatic consolidation should work under a standalone compatible host without requiring a separately installed executable named `pi`. Possible designs include:

- provide the automatic consolidator with extension runtime model/model-registry context so it can use the in-process direct transport; or
- resolve a host-provided child CLI invocation instead of hard-coding Pi-specific discovery.

## Additional context

The manual `/memory-consolidate` path can receive direct runtime context, so this specifically affects automatic consolidation initiated by `MemoryStore` when a store is full.

[...]

PR #216, opened 2026-09-01
feat: opt-in active recall at session start

wtchronos:
## Summary
- Adds opt-in `activeRecallEnabled` (default `false`) so `before_agent_start` can search Pi local SQLite memory against the user prompt and inject at most three source-labeled leads.
- Adds a separate opt-in `activeRecallWcmEnabled` (default `false`) for bounded WCM search, with secret-prompt suppression, query redaction, and secret-bearing result filtering.
- Lookup failures leave startup unchanged; this path never writes memory. 0.9.7 behavior is unchanged until both flags are explicitly enabled.

Sliced from the live 0.9.7 overlay (`~/.pi/agent/npm/node_modules/pi-hermes-memory`) onto `v0.9.7`, then rebased onto current `main` so the CHANGELOG does not conflict with #213. Durability / correction-retry work is out of scope.

## Test plan
- [x] `npx tsx --test tests/handlers/active-recall.test.ts` — 11/11 pass
- [x] `npx tsx --test tests/config.test.ts` — flag defaults, boolean honor, non-boolean ignore
- [x] Recorded mutations go red: disabled early-return skipped; `scanSecrets` WCM suppress skipped; both restored
- [ ] Upstream CI on this PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Opt-in active recall**: `activeRecallEnabled` (default `false`) searches Pi local SQLite memory against the user prompt at `before_agent_start` and injects at most three source-labeled leads. `activeRecallWcmEnabled` (default `false`) is a separate opt-in for bounded WCM search. Lookup failures leave startup unchanged; this path never writes memory.
+
 - **Opt-in memory lifecycle timings**: `PI_TIMING=1` now reports startup synchronization and loading, backfill and live-index callbacks, shutdown flush/index/wait/close, and database open/integrity-check/checkpoint durations. Normal runs produce no timing output.
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


chandra447 (commented):
Adversarial review (4/4 reviewers converged) — 1 blocking issue:

**WCM recall blocks every turn, not once per session** (src/index.ts:248-257, src/handlers/active-recall.ts). `before_agent_start` fires after every user prompt, and the handler awaits `searchWcmMemories` — `execFile(~/.local/bin/wcm)` with an 8s timeout and no AbortSignal. With both flags on, each turn can stall the agent loop up to 8s on a slow/missing WCM, uncancelable. README/CHANGELOG describe session-start recall, which understates the cost. Fix options: run WCM off the critical path (fire-and-forget with late injection or cache), thread the turn abort signal with a much smaller budget, and/or restrict recall to actual session start. Also note the hardcoded `~/.local/bin/wcm` path won't resolve on Windows — consider making it configurable (cf. #211's OMP-on-Windows lesson). Local FTS portion is cheap and correctly fenced/labeled; no issue there.

PR #223, opened 2026-09-05
fix(sessions): silently skip non-session jsonl artifacts during backfill

IstPlayer:
## Symptom

On machines with heavy subagent usage, startup shows a noisy warning and burns the backfill budget:

```
Warning: 🧠 Session backfill complete: 0 indexed, 24 skipped, 0 messages (41 file errors).
```

All 41 "file errors" are `subagent-artifacts/*.jsonl` (pi-subagents worker transcripts) — files that were never sessions.

## Root cause

The sessions root is a **shared namespace**: pi core writes session transcripts there, but extensions also create sibling files/directories with `.jsonl` artifacts. `getSessionFiles()` treats every depth-1 subdirectory as a project dir and every `.jsonl` inside as a session; `parseSessionFile()` returns `null` for non-session content, and `indexChangedSessions()` counts each `null` as a per-file error.

## Fix (three layers, as discussed for the general case)

1. **Content sniffing** — new `isSessionFile()`: a file is a session candidate iff its first line parses as JSON with `type: "session"` (the documented session header). Non-session artifacts are skipped **silently** and reported via a new `nonSessionSkipped` counter appended to the backfill notice.
2. **Error tiering** — files that pass the sniff but fail to parse mid-way are still reported as errors; genuine format corruption is never hidden.
3. **`sessionIndexExclude` config** — list of directory names under the sessions root that are never scanned, as an escape hatch for artifacts that do carry a session header but shouldn't be indexed.

### Boundary note

Fork transcripts (`<session>/forks/*.jsonl`, which *do* start with a valid session header) live at depth 3 and stay outside the scanner's depth-2 walk, so behavior is unchanged for them. Whether forks should be searchable is a separate feature decision, deliberately not taken here.

## Tests

- 21/21 in `tests/store/session-parser.test.ts` (incl. 5 new `isSessionFile` + exclude cases)
- 51/51 in `tests/store/session-indexer.test.ts` (incl. new non-session-skip / error-tiering cases)
- Full suite: **all 52 test files pass**; `tsc --noEmit` clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,6 +496,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
   "projectsMemoryDir": "projects-memory",
   "sessionSearch": { "variant": "legacy" },
   "sessionRetentionDays": 0,
+  "sessionIndexExclude": [],
   "quickCheckOnOpen": true,
   "llmModelOverride": "openrouter/deepseek/deepseek-v4-flash",
   "llmThinkingOverride": "off",
@@ -535,6 +536,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `projectsMemoryDir` | `projects-memory` | Subdirectory under `~/.pi/agent/` for project-scoped memory |
 | `sessionSearch` | `{ "variant": "legacy" }` | Session search implementation: `legacy` keeps the existing SQLite/FTS snippet search; `anchors` uses the opt-in Markdown request surface and returns compact JSONL line-range anchors from `~/.pi/agent/sessions/` |
 | `sessionRetentionDays` | `0` | Opt-in SQLite session retention, in days. `0` (default) disables pruning entirely and keeps the legacy count-only backfill preflight. When positive, sessions whose JSONL source file was last modified longer ago than the window are pruned from SQLite at startup — **rows only; the JSONL files in `~/.pi/agent/sessions/` are never deleted** — and both the deferred backfill and `/memory-index-sessions` skip files outside the window, so pruned sessions stay pruned instead of being re-indexed |
+| `sessionIndexExclude` | `[]` | First-level directory names under the sessions root to exclude from session indexing (exact names or `*` globs). The sessions directory is a shared namespace — extensions may store non-session JSONL artifacts there. Non-session files are already skipped by content sniffing (first line must be a `{"type":"session",...}` header) and no longer surface as backfill errors; use this setting to also skip whole directories regardless of content, e.g. `["subagent-artifacts"]` |
 | `quickCheckOnOpen` | `true` | Run a full SQLite integrity check asynchronously after opening the database; set to `false` to skip the startup scan (operation-time recovery remains enabled) |
 | `llmModelOverride` | unset | Optional model override for background review (direct and subprocess), correction save, session flush, and consolidation |
 | `llmThinkingOverride` | unset | Optional thinking override for those LLM calls; valid values are `off`, `minimal`, `low`, `medium`, `high`, and `xhigh`. If `llmModelOverride` is set and this is omitted, review/child calls default to `off` |
--- a/src/config.ts
+++ b/src/config.ts
@@ -69,6 +69,7 @@ const DEFAULT_CONFIG: MemoryConfig = {
   sessionSearch: { variant: "legacy" },
   quickCheckOnOpen: true,
   sessionRetentionDays: DEFAULT_SESSION_RETENTION_DAYS,
+  sessionIndexExclude: [],
 };
 
 export const DEFAULT_CONFIG_PATH = path.join(
@@ -146,6 +147,7 @@ export function loadConfig(configPath = DEFAULT_CONFIG_PATH): MemoryConfig {
         config.sessionRetentionDays = parsed.sessionRetentionDays;
       }
       if (typeof parsed.standingInstructionsEnabled === "boolean") config.standingInstructionsEnabled = parsed.standingInstructionsEnabled;
+      if (isStringArray(parsed.sessionIndexExclude)) config.sessionIndexExclude = parsed.sessionIndexExclude;
       if (typeof parsed.projectCharLimit === "number") config.projectCharLimit = parsed.projectCharLimit;
       if (typeof parsed.memoryDir === "string") {
         const normalizedMemoryDir = normalizeConfiguredMemoryDir(parsed.memoryDir);
--- a/src/handlers/index-sessions.ts
+++ b/src/handlers/index-sessions.ts
@@ -41,7 +41,7 @@ export function registerIndexSessionsCommand(pi: ExtensionAPI, config: MemoryCon
         try {
           // Retention is honored here too: a manual reindex must not re-add the
           // expired sessions the auto pruning just deleted.
-          const result = indexAllSessions(dbManager, SESSIONS_DIR, undefined, retentionCutoffMs(config.sessionRetentionDays));
+          const result = indexAllSessions(dbManager, SESSIONS_DIR, undefined, retentionCutoffMs(config.sessionRetentionDays), config.sessionIndexExclude);
           const stats = getSessionStats(dbManager);
 
           let output = `\n✅ Session indexing complete!\n\n`;
--- a/src/handlers/session-backfill.ts
+++ b/src/handlers/session-backfill.ts
@@ -32,6 +32,8 @@ export interface ScheduleSessionBackfillOptions {
   needsBackfillFn?: typeof needsBackfill;
   indexSessionsFn?: typeof indexChangedSessions;
   maxFilesToIndex?: number;
+  /** First-level dirs to skip in the all-projects scan (sessionIndexExclude config). */
+  sessionIndexExclude?: string[];
   touchBackfillTimestampFn?: typeof touchBackfillTimestamp;
   /**
    * Optional retention cutoff (ms epoch). Files older than this are excluded
@@ -43,8 +45,9 @@ export interface ScheduleSessionBackfillOptions {
 
 function formatBackfillResult(result: BulkIndexResult): string {
   const errorSuffix = result.errors.length > 0 ? ` (${result.errors.length} file error${result.errors.length === 1 ? '' : 's'})` : '';
+  const nonSessionSuffix = (result.nonSessionSkipped ?? 0) > 0 ? ` (${result.nonSessionSkipped} non-session files skipped)` : '';
   const limitSuffix = result.reachedLimit ? ' (startup limit reached)' : '';
-  return `🧠 Session backfill complete: ${result.sessionsIndexed} indexed, ${result.sessionsSkipped} skipped, ${result.messagesIndexed} messages${errorSuffix}${limitSuffix}.`;
+  return `🧠 Session backfill complete: ${result.sessionsIndexed} indexed, ${result.sessionsSkipped} skipped, ${result.messagesIndexed} messages${errorSuffix}${nonSessionSuffix}${limitSuffix}.`;
 }
 
 function notifyBestEffort(notify: NotifyFn | undefined, message: string, level: NotifyLevel): void {
@@ -102,7 +105,7 @@ export function scheduleSessionBackfill(
     setTimeoutFn(() => {
       measureLifecycleSync('session-backfill.callback', () => {
         try {
-          const result = indexSessionsFn(dbManager, sessionsDir, { maxFilesToIndex, retentionCutoffMs });
+          const result = indexSessionsFn(dbManager, sessionsDir, { maxFilesToIndex, retentionCutoffMs, excludeDirs: options.sessionIndexExclude });
           if (!result.reachedLimit) touchBackfillTimestampFn(dbManager);
           notifyBestEffort(options.notify, formatBackfillResult(result), result.errors.length > 0 || result.reachedLimit ? 'warning' : 'info');
         } catch (err) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -257,6 +257,7 @@ export default function (pi: ExtensionAPI) {
         // Exclude files older than the retention cutoff so sessions pruned by
         // retention are never re-indexed and never schedule another backfill.
         retentionCutoffMs: retentionCutoffMs(config.sessionRetentionDays),
+        sessionIndexExclude: config.sessionIndexExclude,
       });
     }
   });
--- a/src/store/session-indexer.ts
+++ b/src/store/session-indexer.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import { DEFAULT_MAX_MESSAGE_CONTENT_LENGTH } from '../constants.js';
 import { DatabaseManager } from './db.js';
-import { parseSessionFile, getSessionFiles, type ParsedSession } from './session-parser.js';
+import { parseSessionFile, getSessionFiles, isSessionFile, type ParsedSession } from './session-parser.js';
 
 export const LAST_SESSION_BACKFILL_KEY = 'last_session_backfill';
 export const SESSION_BACKFILL_INTERVAL_MS = 24 * 60 * 60 * 1000;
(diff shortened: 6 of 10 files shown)


chandra447:
Hey, thanks for picking this up — genuinely grateful for the contribution toward the extension, this backfill noise has been bugging people on heavy subagent setups and your diagnosis of the shared-namespace root cause is spot on.

I ran the diff through a few reviewer passes and there's one real gap before this can land: `needsBackfill` (and `countSessionFiles`) don't know about the sniff or `excludeDirs`. Skipped/excluded files never get `session_files` rows, but the preflight still enumerates them via unfiltered `getSessionFiles()` and returns true on the first unstamped file — before the 24h timestamp is even checked. So on an artifact-heavy machine the 41-error warning goes away, but every startup still schedules a deferred pass plus the `(... N non-session files skipped)` notice. The scheduling churn this PR is meant to kill survives. Could you thread `excludeDirs` into `needsBackfill`/`countSessionFiles` and the `needsBackfillFn` call in `scheduleSessionBackfill`, plus a regression test (artifacts-only dir + fresh timestamp → `needsBackfill` false)?

Two smaller things worth fixing in the same pass:

- In `indexChangedSessions` the sniff runs before the stat/metadata short-circuit, so the steady-state no-op startup pays open+read per file, and genuinely changed files get sniffed twice (gather loop + `indexSessionFile`). Reordering to stat → stored-match → sniff-only-changed, and sniffing once, keeps the same behavior for less I/O.
- `isSessionFile` fails closed on truncation: if the first line runs past the 4KB window there's no newline in-buffer, `JSON.parse` throws on the fragment, and a real session silently becomes `nonSessionSkipped`. Given the function already returns true for unreadable files, worth returning true when no newline was observed and letting the full parser decide.

Nits you can take or leave: `/memory-index-sessions` pre-count and result output don't account for the new skip classes (says "Found N" then processes fewer, no non-session line unlike the backfill notice), and `sessionIndexExclude` entries aren't trimmed/normalized so `" subagent-artifacts "` silently matches nothing.

Nothing on the Pi extension-contract side — factory stays light, heavy work stays deferred, notify/command shapes look right.

Thanks again for contributing, looking forward to the update.

IstPlayer:
Thanks for the thorough review — all four addressed in 04cedd9:

**1. Preflight gap** — `needsBackfill` and `countSessionFiles` now share the indexer's filtered view (`getSessionFiles(..., excludeDirs, sessionOnly=true)`), and `scheduleSessionBackfill` threads `sessionIndexExclude` into the `needsBackfillFn` call. Regression tests: `needsBackfill is false for an artifacts-only sessions dir with a fresh timestamp` and `needsBackfill is false when only excluded dirs exist and the last backfill is recent` (plus a sanity assertion that the same state *does* demand backfill without the exclusion).

**2. I/O ordering** — `indexChangedSessions` gather loop is now stat → retention → stored-match → sniff, so steady-state (stored-match) files never pay the sniff at all. Changed files are sniffed exactly once: `indexSessionFile` takes an `alreadySniffed` flag and the post-cap loop passes `true`.

**3. Fail-open** — refined rather than blanket: when no newline is observed in the 4KB window we still try to parse the whole window; a *parseable* non-session first line (complete single-line artifact without trailing newline) stays silently skipped, only a genuinely truncated fragment (parse throws) fails open to the full parser. Empty file / blank first line keep the old `false`. Covered by `isSessionFile fails open when the first line exceeds the sniff window (no newline observed)`.

**4. Nits** — `/memory-index-sessions` pre-count goes through the same filtered view (so "Found N" matches what's processed) and appends `(N non-session files will be skipped)` when relevant; the result block gained a `Non-session files skipped` line; `matchesExcludePattern` trims entries so `" subagent-artifacts "` matches.

Full suite green: `tests/run-all.sh` → all 52 test files pass, `tsc --noEmit` clean.

chandra447:
Hey @IstPlayer, thanks for the quick and complete turnaround — both of our verification passes independently confirmed all four items fixed, and the new tests pin the exact failure modes we called out:

1. `needsBackfill` / `countSessionFiles` share the indexer's filtered view, `sessionIndexExclude` reaches both the preflight and the deferred pass, and the artifacts-only + fresh-timestamp regression is in.
2. The gather loop is stat → retention → stored-match → sniff, changed files sniff exactly once via `alreadySniffed`, and steady-state files never pay the sniff.
3. Fail-open only when no newline was observed and the fragment can't parse; parseable complete non-session lines stay silently skipped; empty and blank stay false. The huge-header regression test covers it.
4. `/memory-index-sessions` pre-count and result output account for the skip classes, and exclude patterns are trimmed at match time.

CI is green on `04cedd9` (Lint, Min Supported SDK, Tests, Type Check).

One perf regression to address before merge, and it's on the very path this PR improves:

**`needsBackfill` now sniffs every file on every startup.** With `sessionOnly=true`, the preflight enumerates and `isSessionFile`s every `.jsonl` — including already-indexed sessions — synchronously inside `session_start`, before the cheap count/stamp/timestamp checks get a chance. On main this check is readdir + stat only; now it's O(n) open + 4 KiB read per launch, which is exactly the cost the I/O-ordering fix elsewhere in this PR avoids, and artifact-heavy machines without `sessionIndexExclude` still pay a sniff per artifact every launch. Suggested shape: mirror `indexChangedSessions` in `needsBackfill` — enumerate with `excludeDirs` but `sessionOnly=false`, then stat → stored-match skip → sniff only unstamped files, treating sniffed-out non-sessions as non-work. Your artifacts-only regression test keeps passing under that ordering, the steady-state preflight becomes cheap again, and the preflight still terminates. (The `sessionOnly=true` default on `countSessionFiles` is fine to keep; it's the startup preflight call that shouldn't pay it.)

Two smaller ones, take or leave:

- A file that fails open (first line longer than 4 KiB, no newline in the window) and then fails `parseSessionFile` never gets a `session_files` stamp, so `files.length > indexed.count` keeps the preflight true forever — the scheduling churn survives through a narrower door. Consider stamping or discounting fail-open files in the preflight count.
- `/memory-index-sessions` sniffs everything twice (pre-count pass, then the indexer pass without `alreadySniffed`). User-initiated, so low stakes; one shared classified listing would tidy it.

One merge-order note: #228 rewrites the same `session-indexer.ts` / `session-parser.ts` regions into streaming generators and currently has none of the sniff / `excludeDirs` / `alreadySniffed` handling. This PR should land first, and #228's rebase needs to carry those across, or the artifact-as-error symptom returns on the streaming path.

Thanks again for the careful fix — this is close.

IstPlayer:
Perf regression addressed in c963278, exactly in the suggested shape:

- `needsBackfill` now enumerates with `sessionOnly=false` (exclusions still applied) and per file does stat → stored-match skip → sniff **only unstamped/changed** files; sniffed-out artifacts count as non-work. Steady-state startup is back to readdir + stat + indexed-DB lookups, no open/read per session inside `session_start`.
- The `files.length > indexed.count` shortcut is dropped rather than adapted: with artifacts in the tree it counts files that can never receive stamps — the same permanent-true churn this preflight exists to terminate. The metadata loop is cheap enough to always pay.
- Retention branch: sniff happens only after a stored-match miss; `hasRetainedFile` keeps its raw meaning so the timestamp behavior is unchanged.
- `countSessionFiles` keeps its `sessionOnly=true` default per your note.
- New regression test: five indexed sessions corrupted in place with their stamps re-pointed at the corrupted stat values + fresh timestamp → `needsBackfill` false. If the preflight ever sniffs stamped files again, the corrupted first line fails open and the test goes red.

All previous tests pass under the new ordering, including the artifacts-only + fresh-timestamp case. Full suite: all 52 test files green, `tsc --noEmit` clean.

On the two take-or-leave items: I left both out deliberately. (1) Fail-open files that then fail `parseSessionFile` do keep the preflight true, but a non-session file with a >4 KiB first line is rare enough that stamping/discounting them would add state for a case the exclusion config already covers. (2) The `/memory-index-sessions` double sniff is user-initiated and one-shot, so it doesn't sit on the startup path this PR is about. Happy to revisit either if you disagree.

Noted on #228 — agreed the sniff/excludeDirs/alreadySniffed handling must survive its rebase; happy to review that side once it's up.

chandra447 (changes requested):
Reviewed the head (`c963278`) against the pi.dev extension and session-format docs. The mechanism is the right layer, and three things check out: the sniff follows the documented `{"type":"session"}` header contract, `sessionIndexExclude` is opt-in and default-off, and the perf commit keeps steady-state startup to sniffing only unstamped files. The tests assert behavior rather than implementation.

Three things need work before this merges, and I have one question that decides how much of it matters.

## 1. The sniff is narrower than the parser it gates, and it fails open where Pi fails closed

`parseSessionFile` scans every line for a `type: "session"` entry and tolerates blank and malformed lines along the way. `isSessionFile` judges only the first line of a 4 KB window. Two consequences follow.

A real session whose header line is damaged, or a 0-byte file left by an interrupted write, now returns `false` from the sniff and lands in `nonSessionSkipped` with no error at all. Before this change both surfaced as a backfill error. That is the "genuine format corruption is never hidden" promise from layer 2, and it is the case where silence costs the most, because the file that vanished is the one a user might want to recover.

In the other direction, the catch branch returns `newline === -1`, so a non-session artifact whose first line runs past the 4 KB window is accepted as a session, queued, and then produces the same per-file error this PR exists to remove.

Worth comparing against Pi's own reader, which is the contract this should match. `readSessionHeader` (`node_modules/@earendil-works/pi-coding-agent/dist/core/session-manager.js:290-307`) reads 512 bytes, requires `header.type === "session" && typeof header.id === "string"`, and returns `null` on any parse or I/O failure. It fails closed.

Suggested: require the same two fields and fail closed on parse and I/O errors, treat a 0-byte `.jsonl` as an error rather than an artifact since no legitimate writer produces one, and let the sniff tolerate the same leading-line shape the parser already accepts.

## 2. The preflight does not converge

`needsBackfill` lost the `files.length > indexed.count` short-circuit, so every `session_start` now pays a readdir, a stat, and a `session_files` lookup per file, plus an open and a 4 KB read for every artifact. Artifacts never get a stamp, so nothing about the second launch is cheaper than the first. On an artifact-heavy machine that is a permanent per-launch cost, and it is the cost the original issue was complaining about in a different currency.

Persisting a negative stamp through the same machinery you already use (path plus size plus mtime means not-a-session) would make the second launch readdir, stat, and lookup only. If you would rather not persist anything, measure it and put the number in the commit so the trade is visible.

## 3. The escape hatch is off by default

`sessionIndexExclude` defaults to `[]` (`src/config.ts:72`), so the `subagent-artifacts/` tree from the report is still walked unless the user reads the README and adds it. Either ship the known artifact directories in the default list, or make the skip work without config. Fixing the sniff contract in item 1 covers most of the second option.

## Question: what do the 41 files actually look like?

The new tests fabricate a depth-2 `subagent-artifacts/*.jsonl` containing `{"type":"transcript",...}`. On a live machine the real subagent artifacts sit one level deeper than the walk reaches and are byte-identical in shape to real Pi sessions, starting with a valid session header. If the files in the original report carry a session header, the sniff will not skip them, and the exclusion list being off by default becomes the whole fix.

Could you paste the first line, redacted, and the path of one of the 41 files? That settles whether the sniff or the exclusion is doing the work here.

## Smaller items

- The classification decision now lives in four call sites with three protocols: the `alreadySniffed` flag on `indexSessionFile`, the manual sniffs in both `needsBackfill` branches, and the duplicate counter in the command handler. One helper that returns `unchanged | session | nonSession` would collapse it and give the flag a single owner.
- The 4 KB window can split a multi-byte character at the boundary. Reachability is near zero with current header sizes, but a one-line comment would save the next reader the trace.
- The periodic backfill re-announces the non-session count on every run, which is noise on exactly the machines this fixes.
- The `excludeDirs` threading added in `04cedd9` has no test, and that commit was itself fixing a threading omission.
- An explicit `projectDir` argument bypasses `sessionIndexExclude`.

## Tests to add

- A session file whose first line is damaged is reported as an error, not skipped as an artifact.
- A 0-byte `.jsonl` is reported as an error.
- A first line longer than the window with no newline does not fail open into a session.
- Exclusion threading through `scheduleSessionBackfill` reaches the walk.
- The second launch after a skip costs no file opens.

IstPlayer:
Answered the question with data from this machine (pi-subagents 0.67.0, 23 accumulated agent runs).

**Where they live / what they look like**

```
$HOME/.cache/pi/sessions/subagent-artifacts/<runId>_worker[_N]_transcript.jsonl
$HOME/.cache/pi/sessions/subagent-artifacts/<runId>_worker[_N]_{input.md,output.md,meta.json}
```

93 files = 23 runs × 4. Only the `.jsonl` ones are in the walk (siblings are `.md`/`.json`). They sit **directly under the sessions root (depth 1)**, not one level deeper — I looked for `<sessionId>/subagent-artifacts/` and there is none on this machine.

Redacted first line of `…_worker_0_transcript.jsonl`:

```json
{"version":1,"recordType":"message","source":"async","runId":"<uuid>","agent":"worker","childIndex":0,"cwd":"$HOME","ts":1234567890123,"timestamp":"2026-09-10T13:10:45.962Z","sourceEventType":"initial_prompt","role":"user","text":"…"}
```

There is no `type` field at all — it's a `recordType: "message"` envelope, so the first-line sniff returns false. I also grepped the full files: three transcripts (203–229 lines each) contain **zero** occurrences of `"type":"session"`, i.e. the parser and the sniff agree here; nothing in these files looks like a session to `parseSessionFile` either.

So on this machine **the sniff is what skips them** and the default-off exclusion list is not what's doing the work for this report — the opposite of the case you were guarding against, and it matches the shape of the fixture the tests fabricate. Current inventory: 23 `transcript.jsonl`; zero 0-byte `.jsonl` right now (so I can't quote one, but I agree those should be errors, not silent skips).

**Where I land**

1. Agreed — I'll match Pi's own reader: require `type === "session"` plus a string `id`, fail closed on parse and I/O errors, treat a 0-byte `.jsonl` as an error, and tolerate the same leading-line shape `parseSessionFile` already accepts. Plus the tests you listed.
2. Fair — the short-circuit is gone and artifacts never earn a stamp, so launch two isn't cheaper than launch one. I'd rather persist the negative stamp (path + size + mtime ⇒ not-a-session) through the same `session_files` machinery than reintroduce a count heuristic, and put the measured before/after for an artifact-heavy dir in the commit message so the trade is visible.
3. Given the data above, I don't think shipping known artifact dirs in the default exclusion list is needed for this report — once item 1 lands, the skip works without config. Happy to keep `sessionIndexExclude` opt-in, or add the dirs anyway if you'd rather be belt-and-braces. Your call.

Smaller items: the `unchanged | session | nonSession` helper sounds right (it also gives `alreadySniffed` a single owner), and I'll take the multibyte comment, the repeated non-session count on every periodic run, the missing `excludeDirs` threading test, and the explicit `projectDir` bypass.

Question before I push the rework: is item 3 settled by item 1, and do you want item 2 in this PR or split into a follow-up?

IstPlayer:
bumping this since i think it's stuck waiting on a call from your side, and you asked what the files actually look like.

short version of my 9/17 comment: on this box the artifacts are `~/.cache/pi/sessions/subagent-artifacts/<runId>_worker_0_transcript.jsonl`, 23 of them, sitting at depth 1 under the sessions root so the walk does hit them. the first line is a `{"version":1,"recordType":"message",...}` envelope with no `type` field at all, and grepping the whole files turns up zero `"type":"session"`. so here it really is the sniff that skips them, not the exclusion list — which makes me think item 3 just falls out of fixing item 1, but tell me if you read it differently.

still need your call on the negative stamp: in this pr, or a follow up? item 1 is unambiguous, i'll match `readSessionHeader` (require `type` + a string `id`, fail closed on parse and io, 0-byte `.jsonl` is an error) and add the tests you listed. #233's scope fingerprint is about the shape i'd copy for the stamp, so that part should be uncontroversial.

ci is still 4/4 green and the branch merges clean into main. happy to rebase if you've got something else landing first.

PR #224, opened 2026-09-05
feat: add opt-in lazy memory initialization

Zrzzzz:
## Summary

Add opt-in `lazyInitialization` for `policy-only` memory. Opening Pi currently synchronizes Markdown into SQLite, loads memory stores, and runs maintenance even when no memory operation is needed. This is noticeable when the extension/data live on shared storage.

```json
{
  "memoryMode": "policy-only",
  "lazyInitialization": true
}
```

The default is `false`. `legacy-inject` retains eager loading and its frozen startup snapshot.

## Changes

- Share a retryable first-use initialization across memory tools, data-dependent commands and automatic review/correction/flush operations.
- Keep ordinary prompts, cold shutdown, skill discovery, pinned instructions and anchor-mode session search independent of ordinary memory/SQLite loading.
- Read legacy pinned instructions independently of migration or SQLite. The primary file (even if empty) is authoritative, and pin mutations preserve any loaded legacy instructions.
- First-use readiness includes global initialization, project binding and the complete scheduled backfill pass, with no five-second first-search timeout. Original Pi JSONL files remain untouched, and backlogs larger than the existing 50-file cap still use `/memory-index-sessions`.
- Track preparation and memory tool/command execution through shutdown before closing SQLite; prevent subsequent database reopen. Cancelled tool waiters return promptly without cancelling shared work.
- Keep the direct completion SDK statically imported for Pi's jiti aliases, including production installs without package-local SDK peers.
- Keep project listing SQLite-free; clear failed project bindings for retry, and preserve pending corrections when initialization fails.
- Document the tradeoffs and add an isolated, synthetic-data startup benchmark.

This defers data initialization, not every extension module import. Automatic learning retains its existing triggers and costs; first use pays the deferred initialization cost.

## Validation

- `npm run check`: passed.
- `npm run check:min-sdk`: passed against the declared minimum Pi SDK, 0.80.6.
- `npm test`: 967 tests passed across 55 files; no skipped tests.
- `npm run check:production`: passed. Packs and installs without dev/peer dependencies, verifies package-local SDK resolution is unavailable, loads via Pi's real jiti loader and runs four calls through the shared direct-completion helper against a loopback SSE fixture. No paid model calls or real memories. This is a standalone check; adding it to Actions requires a maintainer or a token with workflow scope.
- Real Pi 0.85.1 TUI startup using a disposable agent directory, explicit local extension and `PI_STARTUP_BENCHMARK=1`: exited successfully, with no memory initialization spans in lazy mode. No model calls or real user memory were used.
- Regression coverage includes cold startup/shutdown, pins despite database failure, skill paths, concurrent first search/write, immediate commands, custom storage, project failure followed by non-project binding/retry, delayed preparation and operation shutdown, no post-close database reopen, legacy migration, eager/legacy snapshots, 40 JSONL files/1000 messages, first-use waiting beyond the shutdown deadline, anchor search, cancellation and correction retry.

The existing frozen-snapshot test used an unawaited `store.add()` plus a 250ms sleep. It also failed on unmodified upstream `34c6fe4` in this environment. This PR replaces the sleep with awaiting the write so the snapshot assertion tests completed persistence deterministically.

## Measurements

Remeasured after restoring the static SDK import and addressing review feedback. Using `node --import tsx scripts/benchmark-memory-startup.mjs [--lazy]`, three sequential eager/lazy pairs, Node 24.19.0, extension checkout/dependencies on GPFS and synthetic data in `/tmp`:

| Metric (median) | Eager | Lazy |
| --- | ---: | ---: |
| Extension startup, including import and registration | 2288ms | 1199ms |
| Session startup hook/discovery | 929ms | 2ms |
| First memory search | 1ms | 1240ms |
| Memory database created during startup | yes | no |

These are extension lifecycle measurements, not full Pi startup or a controlled cold-cache benchmark. Filesystem cache and shared-storage load cause variation. The deterministic regression is that lazy startup does not load ordinary memory or open the memory DB before activation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,6 +487,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 
 ```json
 {
+  "lazyInitialization": false,
   "memoryMode": "policy-only",
   "memoryPolicyStyle": "full",
   "memoryCharLimit": 5000,
@@ -524,6 +525,7 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 
 | Setting | Default | Description |
 |---|---|---|
+| `lazyInitialization` | `false` | Opt in to first-use initialization in `policy-only` mode. Defers Markdown/SQLite sync, ordinary memory loading, maintenance and session indexing until a memory operation needs them. `legacy-inject` keeps eager loading to preserve session snapshots. See below for lifecycle tradeoffs. |
 | `memoryMode` | `policy-only` | Prompt behavior: `policy-only` injects only memory policy; `legacy-inject` restores full memory prompt injection |
 | `memoryPolicyStyle` | `full` | Policy text used in `policy-only` mode: `full` preserves the default v0.7 policy; `compact` uses shorter built-in guidance; `custom` uses `memoryPolicyCustomText`; `none` injects no policy text |
 | `memoryPolicyCustomText` | unset | Custom policy text used when `memoryPolicyStyle` is `custom`; blank or missing text falls back to `compact` |
@@ -562,6 +564,45 @@ Create `~/.pi/agent/hermes-memory-config.json`:
 | `flushMinTurns` | `6` | Minimum turns before flush triggers |
 | `flushRecentMessages` | `0` | Recent messages included in session flush (`0` = all) |
 
+### Optional Lazy Initialization
+
+For installations on slow or shared storage, enable:
+
+```json
+{
+  "memoryMode": "policy-only",
+  "lazyInitialization": true
+}
+```
+
+With this option, opening Pi or sending an ordinary prompt does not initialize
+the memory database or read the ordinary memory stores. Tools and commands are
+still registered immediately. The first memory search, write, or data-dependent
+memory command waits for migration, synchronization and loading. Concurrent
+callers share the load; a failed load can be retried by the next operation.
+
+Important boundaries:
+
+- Pinned `STANDING.md` instructions and skill discovery remain available at
+  startup. If pinned instructions still live in the legacy storage root, that
+  upgrade is performed eagerly so the first prompt cannot silently omit them.
+- `legacy-inject` ignores the lazy option and preserves its startup snapshot.
+- Automatic review, correction capture and flush retain their existing triggers;
+  when a trigger fires, it initializes memory before reading or writing it.
+  Lazy initialization does not disable automatic learning or its model costs.
+- Session indexing starts after memory activation. Until then, Pi's original
+  JSONL session files remain the source of history. The first SQLite
+  `session_search` joins the bounded catch-up pass (at most 50 changed files);
+  use `/memory-index-sessions` for a larger backlog. Anchor-mode session search
+  reads JSONL directly and does not activate the memory database.
+- Closing an unused session does not initialize memory just to index it. A
+  configured flush that meets its minimum-turn threshold can still activate it.
+- This defers data initialization, not every extension module import. The direct
+  completion SDK is also imported on demand, when an LLM-backed operation runs.
+  First use pays the deferred cost; this is not a guarantee of faster searches.
+
+The default remains `false`, so existing installations keep eager initialization.
+
 ## Diagnosing lifecycle latency
 
 Run Pi with timing enabled to see which memory lifecycle step is slow:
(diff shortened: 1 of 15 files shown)


chandra447:
Hey, thanks for this — genuinely grateful for the contribution toward the extension. Lazy init for policy-only is something we've felt on shared storage, and the overall design here is right: shared retryable initializer, legacy-inject staying eager, prompts/shutdown/skills/pins/anchors staying SQLite-free. I ran it through a few reviewer passes and the feature is shippable, but there are five things to fix first, one of them blocking.

The blocker: the `completeSimple` change to `import type` + dynamic `import()` inside `completeOnce` breaks both `tsc` (TS1361 on `typeof` against a type-only name) and production runtime — native dynamic import defeats Pi's jiti aliasing, and since `@earendil-works/pi-ai` is peer/dev-only, direct-transport review/flush/correction/consolidate throws on first use under a real `pi install --omit=dev`. Please restore the static value import (`review-memory-ops.ts:6, :557`).

The other four:

- `/memory-switch-project` only `readdir`s Markdown via `fs` and never touches stores/SQLite, so registering it on `memoryPi` pays full migration + load + backfill for a listing on a cold lazy session. Register it on raw `pi` instead (`index.ts:357`).
- First-use readiness is split: `isReady()` flips at the end of global init while project bind and the search backfill wait still run, so shutdown can `close()` under in-flight first use (and `getDb()` later reopens past the WAL checkpoint), and the first SQLite search proceeds against a partial index — on exactly the slow storage this feature targets. Treat first-use as init + bind + backfill wait and join all of it before shutdown `close()`; the 5s shutdown cap shouldn't double as a first-search join (`index.ts:265-267, :373-398`; `memory-initialization.ts:40-43`).
- A failed project `loadFromDisk` resets the dir to null but leaves `projectLoad` rejected, so the next bind re-awaits the stale failure and poisons every later bind. Clear it on failure (`index.ts:414-421`).
- The legacy `STANDING.md` upgrade is tied to `ensureMemoryReady`, which rethrows on lazy failure, so `session_start` aborts before `standingStore.load()` and the first prompt silently omits pins — the thing that branch exists to preserve. Load standing pins independently of full init (`index.ts:269-278`).

Smaller, take or leave: the correction rate-limit is consumed before the init attempt, so a failed lazy load swallows a user correction with no retry (flush already returns early on init failure — correction should match); and Escape during the session-search backfill wait can't interrupt since `signal.throwIfAborted()` only brackets the whole `ensure`.

Nothing on the Pi extension-contract side — deferral points match the docs, no background-resource or event-shape issues. The lazy tests cover the happy paths well; worth noting the backfill test uses one tiny JSONL so it can't catch the partial-index path behind item three — a bigger fixture would pin it.

Thanks again for contributing, looking forward to the update.

Zrzzzz:
Thanks for the detailed review. I've addressed all five requested changes, plus the two smaller points:

1. Restored the static `completeSimple` value import and removed its call-time dynamic import. Added `npm run check:production`. It packs/installs with `--omit=dev --omit=peer`, verifies SDK peers cannot resolve beside the package, then uses Pi's real jiti loader and a loopback SSE endpoint to exercise the direct-completion path without real credentials or paid model calls.
2. Registered `/memory-switch-project` on raw `pi`. The cold-start regression now invokes the listing and confirms no memory-store load or memory database open.
3. Readiness now includes global initialization, project bind and the scheduled backfill's completion. An uncapped first-use join replaces the five-second shutdown wait. The lifecycle also tracks tool/command execution, and shutdown joins that work before checkpoint/close; a retained open guard rejects late database reopen. Escape cancels the individual waiting tool promptly while underlying shared preparation remains owned and is joined at shutdown.
4. A failed project load clears the rejected promise and partial binding. Added a regression for failure followed by a non-project cwd and then a successful project retry.
5. `StandingInstructions` reads legacy pins independently of full memory initialization. The primary file, including an intentionally empty file, wins. First pin writes carry forward loaded legacy instructions. A regression forces database initialization failure and verifies pins remain in the first and subsequent prompts and skills remain discoverable.

Correction initialization failures now retain the original correction text and do not consume its rate limit, so the next eligible turn can retry even after an ordinary follow-up message.

Validation: `npm run check`, `npm run check:min-sdk` (0.80.6), `npm run check:production`, and all **967 tests / 55 files** passed locally without skips. Additional coverage includes a 40-file/1000-message history fixture, a join held past the five-second shutdown deadline, delayed project binding, shutdown during execution, cancellation and post-close reopen rejection. The existing 50-file backfill cap is unchanged: the entire scheduled pass is joined, while larger backlogs still use `/memory-index-sessions` as documented.

Also reran real Pi 0.85.1 startup with an isolated agent root and legacy pins: it exits cleanly with no lazy memory-init spans. Updated the PR description and measurements to reflect the restored static SDK import and the new lifecycle.

One permission note: GitHub rejected the push when it included adding the production smoke step to Actions because my OAuth authorization lacks `workflow` scope. I removed that workflow-only change rather than broadening permissions. The standalone script is included and passed locally; it can be added after `npm test` in CI by a maintainer.

chandra447:
Hey @Zrzzzz — verified on this end: two independent verification passes confirmed all five requested changes plus the two smaller ones, with file:line evidence on each. This is ready to merge from our side.

1. Static `completeSimple` value import restored (`review-memory-ops.ts:7` and the single binding at `:274`). `check:production` is a genuinely good addition — it packs with `--omit=dev --omit=peer`, asserts the SDK peers are unresolvable beside the package, then loads through Pi's real jiti `loadExtensions` and drives four loopback completions. That is precisely the regression class the dynamic-import version had.
2. `/memory-switch-project` on raw `pi`, with the cold-start regression asserting zero store loads and zero database opens.
3. Readiness = global init + project bind + uncapped `joinSessionBackfill`; `close()` joins tracked preparation and in-flight tool/command execution; the open guard rejects post-close `getDb()`; Escape rejects only the waiter while shared preparation stays owned and joined at shutdown. The bind-held-past-5s and 40-file/1000-message fixtures pin this well.
4. Failed `loadFromDisk` clears the rejected promise and partial binding; failure → non-project cwd → successful retry is covered.
5. Pins load before any init in `session_start`, the primary file is authoritative even when empty, legacy text carries into the first primary write, `EISDIR` is not masked by the fallback, and the database-unavailable regression proves pins and skills survive.
6. Correction rate limit is preserved on init failure and the text is carried forward.
7. Escape interrupts a waiting tool during the first-use wait without touching shared preparation.

Eager mode is confirmed unchanged (the backfill join stays lazy-gated, eager shutdown keeps its 5s caps, and there is no double-close between the unused and used shutdown paths). The lifecycle redesign — tracked preparation, uncapped first-use join, open guard — is the right shape, and the tests defend it rather than the implementation.

Three nits, none blocking:

- A correction that arrives during the 3-turn cooldown is still dropped (pre-existing shape — `pendingCorrection` clears before the rate-limit check). If you want the same retry semantics as the init-failure path, restore it with `pendingCorrection ??= correctionText` on that return.
- The correction init-failure catch treats an agent abort like a transient load failure and restores the pending text. Fine either way; worth a deliberate thought on whether Escape should drop the attempt.
- The `check:production` probe ignores its four operation labels — every iteration runs the same `runDirectMemoryCompletion`. The smoke's real assertion is the jiti import path, so the loop could be a single call, or the operations could actually vary.



Great work — thanks for the patience through two review rounds and for the thorough validation (the production install check especially).

PR #228, opened 2026-09-09
fix: stream and incrementally checkpoint session indexing

Noir-Lime:
## Problem

Session indexing currently reads the entire JSONL file into a JavaScript string before splitting and parsing it. On a large, media-heavy session this can exceed V8's maximum string length or exhaust the heap. Live indexing repeats that allocation as messages finish.

Observed failures included `Cannot create a string longer than 0x1fffffe8 characters` and a V8 heap OOM in a long-lived Pi Web session daemon. The release and upstream main checked at implementation time still used the whole-file reader.

## Changes

- Tokenize JSONL in 64 KiB chunks with stream-json, without assembling whole files or individual records.
- Retain bounded searchable text and tool names while discarding media payloads, reasoning, and tool arguments/output during parsing.
- Write bounded batches (64 messages or 512 KiB of text).
- Persist byte checkpoints in extension_metadata, validating file identity, size/mtime, and prefix/boundary fingerprints before resuming an append.
- Retry incomplete trailing records and rescan on detected truncation/replacement or missing indexed sessions.
- Route live, backfill, manual, and shutdown indexing through cooperative asynchronous scans; retain synchronous APIs for compatibility.
- Coalesce live-index requests without losing another session or a message arriving during an active scan.
- Add regression tests, implementation notes, and a privacy-preserving benchmark script.

Session source files and Markdown memories are not rewritten. This does not change Pi core's session format or its in-memory history representation.

## Validation

- `npm run check` passed.
- `npm test`: all 53 test files passed.
- Focused regression run after adding the live-index concurrency test: 60 tests passed.
- Installed extension loaded successfully through Pi's extension loader.
- An actual ~1.32 GB session indexed into a temporary database using a 128 MiB V8 heap limit:
  - First scan: ~7.4 seconds; 2,058 searchable messages.
  - Unchanged repeat: <1 ms.
  - Peak process RSS: ~198 MiB (includes native/runtime overhead).
  - Timers continued firing during the scan; maximum observed timer interval ~45 ms.

Reproduce against a local session without changing its production index:

```sh
node --max-old-space-size=128 --import tsx scripts/benchmark-session-index.ts /path/to/session.jsonl
```

## Scope / Limitations

- Checkpoints assume Pi's append-only JSONL contract. An arbitrary in-place edit outside the fingerprint samples combined with an append is not detectable.
- This fixes the indexing path; it does not redesign the optional raw-file anchor-search implementation.
- The synchronous compatibility parser still returns an array of parsed messages; production indexing uses bounded batches instead.


chandra447 (changes requested):
Hey @Noir-Lime, thanks for this — it's a serious piece of work aimed at a real production failure we've seen ourselves (the `Cannot create a string longer than 0x1fffffe8` crash and heap OOM on media-heavy sessions). The streaming design is the right call. `stream-json` is pure JS, BSD-3, no native addons and no install scripts, so it survives `pi install --omit=dev`, and hand-rolling an incremental JSON tokenizer would re-litigate escape handling, `\uXXXX`, and split UTF-8 for no gain. The 1.32 GB / 7.4s / ~198 MiB RSS measurement with TUI timers still firing is exactly the evidence we wanted, and the checkpoint validation (ino/dev/size/mtime, first and boundary 4 KiB fingerprints, orphan-session check) is thoughtfully layered. We ran the diff through four independent adversarial review passes. Two things need fixing before merge, and a few smaller notes in the same pass.

## 1. A checkpoint is thrown away whenever the file grows during the scan (blocker)

`indexFileSteps` publishes the cursor only when `after.size === stat.size && after.mtimeMs === stat.mtimeMs`, with `stat` captured at scan start and `readSessionRecords` freezing `end = fstatSync(fd).size`. On the exact scenario this PR exists for — a first index of a multi-GB session while the user keeps chatting — any append during the 7.4s pass changes size/mtime, the `INSERT INTO extension_metadata` is skipped, and the offset progress is discarded. The next live, shutdown, or backfill pass finds no valid cursor and re-tokenizes the whole file from byte 0. The fixtures are static between `indexLiveSessionAsync` calls, so the suite cannot catch this.

Since the cursor is already validated via a prefix fingerprint at `offset`, publishing `{offset: processed, size: processed, fingerprint: fingerprint(file, processed)}` whenever ino/dev are unchanged is safe: an append leaves the prefix hash valid, a replacement invalidates it. That also fixes a second complaint: checkpoints currently publish only after the entire pass, so a crash at 90% of a first import replays from 0. Publishing after each committed batch (still gated on ino/dev) bounds replay cost too.

## 2. Shutdown is no longer bounded (blocker)

The new `session_shutdown` handler does `await measureLifecycle("shutdown.active-index", () => indexLiveSessionAsync(...))`. `indexLiveSessionAsync` chains onto the module-level `indexingQueue`, so quitting now waits for any in-flight 50-file startup backfill plus the current-file scan. All four review passes flagged this independently. The existing 5s `waitForSessionBackfill` / `waitForLiveSessionIndex` only run after the unbounded await, so they cannot cap it. The old code was also unbounded (worse — it parsed whole files synchronously), so this is the moment to fix the class: race the shutdown scan against `SESSION_LIVE_INDEX_SHUTDOWN_TIMEOUT_MS`, abort the generator through `drainAsync`'s `steps.return()` on timeout, and let the existing 5s waits remain the actual ceiling before `close()`.

Related window in the same path: with scans now async, the 5s wait expiring can run `dbManager.close()` while a coalesced `drainAsync` is between queued items, and the next `indexSessionOnce` throws into a closed handle (session silently unindexed). Making close mutually exclusive with the queue, or no-oping `indexFileSteps` once the DB is closed, closes that window.

## 3. Smaller correctness notes

- **Async fallback dropped.** `indexLiveSessionOnce` (sync) falls back to `indexCurrentSession` when `indexFileSteps` returns null (file exists but no parseable session header yet); `indexLiveSessionAsync` returns null instead. A file that exists without its header yet indexes nothing on the live path. Keeping the fallback in both restores parity.
- **Sync wrapper is a corruption trap.** `dbManager.withCorruptionRecovery(() => indexLiveSessionAsync(...))` in the pending-map `run()` is a sync try/catch around a promise, so an async corruption rejection bypasses it. Production is saved by `queuedIndex`'s own retry, which makes the outer wrap dead code today and a trap for any injected async `indexLiveSessionFn` (the existing test throws synchronously, so it can't catch this). Drop the outer wrap or make recovery await-aware.
- **Sync `drain()` leaks the fd on throw.** `drain` has no `try/finally steps.return()`, so a mid-pass failure leaves `readSessionRecords`' fd open until GC, and the sync exports sit outside the queue, so a sync caller can interleave with a yielded async scan on the same connection and checkpoint key. Production hooks are all async now — simplest is routing the sync wrappers through the same mutex or trimming them to test-only status.
- **Replacement with a new session id leaves orphan rows.** The `DELETE FROM messages` on an invalid cursor only fires when `cursor.header?.id === entry.id`; a rename-replacement with a new header id indexes the new session while the old `sessions`/`messages` rows stay. (Pre-existing on main, but worth deleting by the old cursor header id regardless of id match while you're here.)

Nits, take or leave: the `replacedMessages` subtraction makes an N-for-N rebuild report `messagesIndexed: 0`; `frames` has no depth cap (a pathological deeply-nested `toolCall.arguments` grows the stack — a cap of a few hundred treated as `invalid` keeps it bounded); the orphan-cursor check looks only at the `sessions` row, so a messages-only wipe never rescans; `parseSessionFile` still accumulates every message (fine now that production indexing doesn't call it — worth a docstring saying so); `session-index-v1:*` checkpoint rows are never cleaned when `pruneOldSessions` removes a session, worth deleting in the same prune transaction.

Test gaps worth one small fixture each: a checkpoint withheld when the file grows during a pass (would have caught item 1), a complete final record without a trailing newline, a same-size rewrite with changed mtime, and a UTF-8 code unit split exactly at the 64 KiB read boundary (the current `é` fixture can pass without exercising a split).

On the Pi contract side there is nothing to fix. No new event shapes, the factory starts no scanners or timers, `message_end` still only schedules the 50ms deferred scan, and the 8ms cooperative yields keep the TUI responsive (your ~45ms max timer interval measurement is good). The coalescing test covering A-running + B + A-again is exactly the regression we wanted.

Sequencing note: #223 rewrites the same `session-indexer.ts` / `session-parser.ts` regions. Land #223 first, then rebase this onto it so the streaming path keeps the sniff / `excludeDirs` / `alreadySniffed` handling — landing #228 without the port would resurface the artifact-as-error symptom on the streaming indexer.

Thanks again for the rigor here — the tokenizer, batching, and replay-safety all held up under adversarial review. The two fixes above are what stand between this and merge.

chandra447 (changes requested):
Reviewed the head (`d27b908`). The streaming parser itself is solid work: bounded per-record memory, byte-level resume, idempotent replay, and genuinely behavioral tests. The reported OOM is real and the core fix is right. What needs work is the integration between the new scanner and everything around it, and three of the four items are cheap.

## 1. Shutdown now blocks on the unbounded index queue

`session_shutdown` awaits `indexLiveSessionAsync`, which joins the new process-wide `indexingQueue` with no timeout. Pi awaits `session_shutdown` in `teardownCurrent` before replacing the session, so quit and session switch now wait on in-flight backfill and giant-file scans. The existing 5-second backfill and live waits never even run until the queue drains. The workload this PR exists to survive is the workload that makes quitting slow.

Bound the shutdown wait. If a scan has not finished by the timeout, persist state and let the next launch resume from the checkpoint. That is what the checkpoints are for.

## 2. An append during a scan throws away the whole pass's checkpoint

`indexFileSteps` publishes the checkpoint only when the file's size and mtime at publish time still match the state the scan started from. Live indexing guarantees the opposite: the file grows while the scan runs. So on an active multi-GB session every pass ends with the checkpoint discarded and the next pass starts from byte 0, forever. That is the "appends only scan new bytes" property failing in exactly the case it was written for.

On growth, keep the checkpoint for the prefix the scan actually completed and handle the tail separately, rather than dropping the progress wholesale.

## 3. Corruption recovery is a no-op on the async path

`withCorruptionRecovery` is a synchronous wrapper but is now handed an async thunk, so a corrupted row discovered mid-scan never triggers the rebuild it was written for. This is the production path.

## 4. Checkpoint rows are never reclaimed

`session-index-v1:*` rows in `extension_metadata` accumulate forever: not pruned on file deletion, not invalidated on replacement. Add them to the existing maintenance sweep.

## Minimality

- The retained synchronous APIs bypass the queue, have no production caller, and can still block the event loop from an async path. Delete them or document a caller.
- `stream-json` is loaded at every extension start, including for sessions that never approach a large file. Lazy-load it on first use.

## Smaller items

- Re-queued live scans skip the persist delay, so a fast message stream re-enters the scan immediately.
- A repeated same-id session header deletes rows indexed earlier in the same scan.
- File replacement with a new session id on the same path orphans the old session's rows.
- A permanently partial trailing record re-triggers backfill on every startup.
- The "512 KiB of text" bound is 512 K UTF-16 code units, not bytes.

## Tests to add

- Shutdown during an active backfill returns within the bounded wait.
- A file that grows during a scan publishes a prefix checkpoint, and the next scan resumes from the prefix instead of byte 0.
- A corrupted row on the async live path triggers recovery.
- Deleting a session file removes or invalidates its checkpoint row.
- An unbalanced trailing record does not re-trigger backfill on the second startup.

Issue #195, opened 2026-08-23
Compatibility report: persistent memory tools now run on DeepSeek Harness via pi2dsh


weijiafu14:
## Compatibility result

The persistent-memory tool path of `pi-hermes-memory` now runs unmodified inside DeepSeek Harness through [pi2dsh](https://github.com/weijiafu14/pi2dsh).

```sh
dsh plugin --profile web add pi2dsh
dsh plugin --profile web add pi-hermes-memory
```

## End-to-end boundary exercised

- process/session A stored a fact with `memory_add`;
- the data was persisted by the package;
- a completely fresh process/session B recovered it with `memory_search`;
- `memory_replace`, `memory_remove`, `session_search` and package commands register through the same host bridge;
- the memory content stayed package-owned rather than being copied into a second DSH memory implementation.

This report is intentionally scoped to the persistent tool/command path that was actually exercised; it does not claim every optional dynamically discovered skill surface has been certified.

Evidence:
<https://github.com/weijiafu14/pi2dsh/blob/main/docs/posting-kit/support-matrix.md#plugins-verified-end-to-end>

This is an interoperability/scope-extension report, not a request to add DSH-specific code here. If documenting DeepSeek Harness as another host is useful, I would be happy to send a README PR. Compatibility issues and contributions are welcome at <https://github.com/weijiafu14/pi2dsh>.

chandra447:
Thanks for this — genuinely glad to see the persistent memory path running end-to-end on DeepSeek Harness via pi2dsh. The boundary you drew is exactly the right discipline for a compatibility report: the persistent tool/command path exercised across two fresh processes (`memory_add` in session A recovered by `memory_search` in session B, plus `memory_replace`, `memory_remove`, `session_search`, and the commands through the same host bridge), with the optional dynamically discovered skill surfaces explicitly marked as not certified rather than silently claimed.

The detail we care about most is also the one you called out: memory content staying package-owned rather than copied into a second DSH memory implementation. That's the interoperability property that matters — one source of truth, host-bridged.

And agreed on scope: no DSH-specific code needed here. The tool/command path running unmodified through the host bridge is the compatibility story; anything host-specific belongs on the host side.

Yes, please do send the README PR documenting DeepSeek Harness as a host — happy to review and merge it. Thanks for taking the time to verify and report this properly.

PR #232, opened 2026-09-10
docs: add DeepSeek Harness persistent memory setup

weijiafu14:
Adds a DeepSeek Harness installation section and a two-session persistent-memory walkthrough, following the invitation in #195. Includes pnpm build approval, update/delete tools, the package-owned storage location, and the headless profile option.

Validated on 2026-09-10 with the published packages `@deepseek-ai/dsh@0.1.5-rc.1`, `pi2dsh@0.24.0`, and `pi-hermes-memory@0.9.8`, using an isolated DSH home and real DeepSeek model calls:

- CLI: add a random fact, recover it in a fresh process, replace it, recover the replacement in another process, delete it, and confirm absence after another restart.
- Web: save a random fact, recall it in a new session whose user input does not contain the value, and remove it through the plugin tools.
- Web commands: `/memory-insights`, `/memory-sync-markdown`, `/memory-pin`, and `/memory-preview-context` complete visibly; a pinned instruction reaches the next session's model context.

The README examples follow those persistent-memory flows. This change edits only `README.md`; `git diff --check` passes.


chandra447 (approved):
Thanks for following up on the #195 invitation with this — it's exactly the right shape for a host-documentation PR: one file, a focused section, and a walkthrough that mirrors the persistent-memory contract (write in one session, recover in a fresh one, replace, delete) rather than a feature list. I verified the claims I can check from here:

- Placement and structure fit the Installation section (sits between the local-test block and the Homebrew/ABI note, alongside the other `###` subsections).
- The multi-package single-add form matches pi2dsh's own walkthrough (`dsh plugin --profile pi-tui add pi2dsh pi-mcp-adapter ...`), and the `web`/`headless` profile names match the built-in surface bundles pi2dsh documents.
- The storage location and the `-w` add flag aren't in pi2dsh's or dsh's READMEs that I can see, so I'm taking those from your same-day validation against `@deepseek-ai/dsh@0.1.5-rc.1` / `pi2dsh@0.24.0` / `pi-hermes-memory@0.9.8` — which is the right provenance for a bridge that iterates quickly.

Two small suggestions, take or leave:

1. **The restart requirement deserves an unconditional mention.** pi2dsh's README is emphatic that "plugins mount at startup" and to restart `dsh` after an add. Your walkthrough only mentions restarting in the approve-builds branch, so a reader who already has `dsh web` running on another profile could follow the happy path and wonder why no tools appear. One clause after the install block would do: "(restart `dsh web` if it was already running — plugins mount at startup)".
2. **Pin the validated dsh version inline.** Since dsh is in developer preview with compatibility-breaking changes expected (their README's own words), one line like "Validated with `@deepseek-ai/dsh` 0.1.5-rc.1, `pi2dsh` 0.24.0, `pi-hermes-memory` 0.9.8 (2026-09-10)" protects future readers when a newer dsh renames the flag or the approval flow. It also documents the `-w` flag's provenance, which is currently only visible in the PR body.

Neither blocks: the section is accurate where I can verify it, and the walkthrough is the honest kind (two sessions, fresh process, delete confirmed absent). Approving so it can go in — and if you'd like the two tweaks above in this PR, I'll re-verify and merge either way.

Issue #227, opened 2026-09-08
session-start.persistence-sync runs an unconditional full markdown reconcile on every startup (2-7s measured)


wweir:
## Summary

`session-start.persistence-sync` dominates interactive startup latency. Measured with a pty over 5 runs with `PI_TIMING=1`:

- banner → loaded: **p50 3.6s, p90 5.6s**
- `session-start.persistence-sync`: **2.1–7.2s per run** (dominant)
- `database.quick-check` on the 71MB `sessions.db`: 0.6–1.7s
- `session-start.load`: 0.65–1.7s (inflated while sync/quick-check contend)

## Root cause

`syncMarkdownMemoriesToSqlite()` (src/handlers/sync-markdown-memories.ts) runs an **unconditional full reconcile on every startup**:

- Parses all 3 global markdown files (`failures.md` at 503 entries in my store, `MEMORY.md`, `USER.md`)
- `scanProjectDirs` walks all project memory dirs (~28 here) and reconciles every project `MEMORY.md`
- Per-entry SQL upsert + full-table SELECT/DELETE reconcile
- No mtime/hash short-circuit exists (there is no skip logic in the file); `persistenceInitialized` only guards against repeats within one process

## Suggested fix

Persist a sync-state table `(file_path, mtime_ms, size, hash)`:

1. At the top of `syncMarkdownMemoriesToSqlite`, compare each file against recorded state
2. If **all** files are unchanged, skip the whole reconcile including `scanProjectDirs`
3. If some changed, reconcile only those files and update their state entries

This keeps the existing `await` semantics in `session_start` intact (no consistency change for `before_agent_start` or the memory tools), while reducing the common startup path to a few stat/hash calls.

Secondary option: `quickCheckOnOpen: false` already mitigates the quick-check cost, but the reconcile is not configurable.

## Environment

- pi-hermes-memory latest (npm), pi 0.85.1, Linux
- Store: 503-entry failures.md, 28 project memory dirs, 71MB sessions.db

Happy to test a patch; the pty + `PI_TIMING=1` harness from the report is reusable.

chandra447:
Thanks for this — confirmed on main, and the numbers match the code shape exactly.

[...]

PR #233, opened 2026-09-10
perf: skip unchanged markdown scopes in sqlite reconcile

chandra447:
## Problem

`session-start.persistence-sync` runs an unconditional full markdown→SQLite reconcile on every startup: `readFileSync` + parse of every file, per-entry upserts, a full-scope SELECT, and orphan DELETEs per scope (`reconcileMarkdownMemoryScope`). Measured 2.1–7.2s per startup on a 503-entry `failures.md` across 28 project dirs against a 71MB `sessions.db` (#227), regardless of whether anything changed.

## Changes

- **Per-scope fingerprints inside the reconcile transaction.** `reconcileMarkdownMemoryScope` records `mdsync:v1:[target, project]` in `extension_metadata` (already home for the session-index checkpoints and FTS tokenizer version — no new table, no schema migration) **in the same `db.transaction` as the upserts and orphan deletes**, only after a successful non-degraded reconcile. A scope is skipped iff a parseable state row exists AND `sha256(JSON.stringify(rawEntries))` matches AND a live `COUNT(*)` of the scope's rows matches `entryCount`. Skip returns the same shape the second-run idempotent pass already produces, so the result contract and command output are unchanged.
- **The skip gate lives in the store, not the walker** — deliberate, for two reasons. The store has four reconcile callers (startup sync, the live mutation observer, the post-failed-op path, the repair command); with the fingerprint inside the shared reconcile, every caller maintains state for free, so a session that edits `failures.md` doesn't re-pay the 503-row sweep at next startup. And fingerprint + rows commit atomically, so a crash can never leave "fingerprint matches, rows don't" state.
- **COUNT guard for the empty-mirror case.** Corruption recovery copies `extension_metadata` wholesale but `memories` separately, so a copied fingerprint over an empty/partial mirror must not skip — the live `COUNT(*)` is half of the skip predicate for exactly that.
- **Strict state validation.** A state value must parse to exactly `{sha256: hex, entryCount: integer}` or it is treated as a miss (reconcile + rewrite).
- **No mtime/size oracle.** Coarse-mtime filesystems and same-size rewrites would skip a needed reconcile under a stat fast-path; hashing the already-in-memory entries is cheaper than one FTS trigger row.
- **`/memory-sync-markdown` is the force path** (always passes `force: true`) and remains the repair for sqlite-only drift that a COUNT-preserving edit would hide from the skip. New and deleted project dirs stay with the existing walk and `mirroredProjects` union, both unchanged.
- Lazy init composes for free: whoever calls `syncMarkdownMemoriesToSqlite` (eager `session_start` or deferred first use) populates fingerprints as a side effect of a successful reconcile.

## Validation

- `npx tsc --noEmit`: clean.
- Targeted: `tests/store/sqlite-memory-store.test.ts` (65 pass), `tests/handlers/sync-markdown-memories.test.ts` (20 pass), `tests/integration/corruption-recovery-integration.test.ts` (2 pass).
- Full suite: **all 52 test files passed** on this branch.
- New coverage asserts the behavior, not the mechanism: a SQL spy throws on any `memories` write on the second unchanged call; a missing fingerprint row must NOT skip (falsifying test for the skip-forever bug class); corrupt state JSON reconciles and rewrites; a same-size/same-mtime rewrite still reconciles; the live-observer composition means a session that wrote memories produces zero inserts at next startup; force repairs COUNT-preserving sqlite-only drift.

Deliberately out of scope: a `filesSkipped` counter for the command output and a stale-key sweep (orphaned keys are harmless — the COUNT guard turns them into a miss — and can be added later if the metadata table ever shows accumulation).


chandra447:
Ran the full adversarial interrogation on this PR (4 independent reviewers). Verdict: no criticals, no breaking defects — the skip predicate, observer/startup hash parity, key injectivity, transaction placement, and corruption interplay all verified sound across all four passes. Three findings were real enough to fix before merge, all now on the branch (commit `8d75f48`):

1. **Copied fingerprints could preserve coerced drift (sharpest find).** Corruption recovery copies `extension_metadata` wholesale and coerces invalid memories values instead of dropping rows — a copied fingerprint could agree with coerced rows while content drifted, and the skip would preserve that indefinitely. `copyRecoverableRows` now strips copied `mdsync:v1:*` fingerprints; a rebuilt database re-mirrors markdown exactly once, at zero cost when healthy. The recovery integration test documents the guarantee.
2. **Every empty scope paid a no-op write transaction per startup (2/4).** Emptied scopes delete their fingerprint, so the skip could never fire for them — a write txn per empty scope per launch, briefly taking the WAL writer lock the live indexer needs. Fixed two ways: the fingerprint delete only runs when a fingerprint actually existed, and the live `COUNT(*)` is only computed after a hash match, so a stateless scope costs two reads.
3. **The two handler-level skip tests couldn't fail on a broken skip (3/4).** Their blocked-write errors were swallowed into `counters.warnings`, so the tests passed even with the feature removed. Both now assert `warnings.length === 0` and the expected `skipped` count.

Accepted tradeoff (documented, not hidden): startup no longer self-heals COUNT-preserving SQLite-side drift — a same-count content swap is skipped at startup and repaired by `/memory-sync-markdown` (the command test pins exactly that). FTS index health is likewise outside the fingerprint; all in-repo writers maintain FTS atomically via triggers, so the residual risk is external tampering, which remains the repair command's job.

Full suite: 52/52 files, tsc clean.

chandra447:
Follow-up review fixes pushed (`222298c`).

- The fingerprint key prefix is now one exported constant (`MDSYNC_METADATA_KEY_PREFIX` in `src/constants.ts`) shared by the store and the corruption-rebuild strip. The strip held its own copy of the versioned literal, so a future prefix bump would have silently stopped stripping and reopened the coerced-row skip hole.
- `reconcileMarkdownMemoryScope` reads the fingerprint row on the force path too, so an emptied scope deletes its stale metadata row instead of leaving it behind.
- The skip boundary is now documented at the predicate and in the CHANGELOG: liveness is a row count, so mirror drift that preserves the count (a hand-edited row, or a search index that lost rows) waits for the forced repair. New test proves `/memory-sync-markdown` re-indexes a scope that lost a `memory_fts` row, which is the case the count guard cannot detect.

PR #234, opened 2026-09-10
fix: bound the compact flush to one shared window

### LINKED ISSUE #225: Compact-time memory flush times out after 30s on long conversations
The flush that runs before compaction (session-flush.ts:142) has a hardcoded 30 second timeout. On a long conversation with a local model, that's often not enough time to even finish processing the prompt, so the flush gets cut off before it can save anything.

Saw this happen twice in a row in my local llama.cpp logs: one attempt got through only 10.5k tokens (cold cache) before being canceled at ~30s, then a retry got through 43k tokens (warm cache) before also getting cut off at the same mark. Both failures are swallowed silently, so compaction itself still finishes fine, just no memory gets saved, and the GPU redoes the same large prompt for nothing each time.

Would it be possible to give this path a longer default timeout, or make it configurable? It seems like the one flush call that can afford to wait a bit longer than the others.

Version 0.9.8, still the same on main.

chandra447:
## Problem

`session_before_compact` awaits the flush before compaction proceeds. `flush()` takes one `timeoutMs` and hands that same number to **both** transports in sequence: direct (`runDirectMemoryCompletion`) then, on any non-ok result or throw, `execChildPrompt` with the same `timeoutMs`. Direct's internal timer aborts its own controller without touching the session signal, so a spent budget is treated as "try the other transport" — the same large prompt is re-sent to the GPU and pays a second 30s + 5s watchdog. The compact catch swallows everything silently. Worst case ~60s on the compaction path, no memory saved, no signal (#225: a local llama.cpp model got cut at 30s twice in a row).

## Changes

- **One remaining-budget window for the compact flush.** `flush()` gets a single wall-clock budget (`flushCompactTimeoutMs`, default 60000ms, raised from the hardcoded 30000). Direct runs first; the `pi -p` fallback receives only the **remainder**, never a second copy of the budget. `resolveFlushHandoff` (`{skip:"session_aborted"} | {skip:"budget_exhausted"} | {timeoutMs}`) is the single decision point at each handoff, so the skip reason is a type rather than scattered conditionals.
- **`flushCompactTimeoutMs` config key** (types, DEFAULT_CONFIG, loadConfig). Honored verbatim; values below 60000 warn at startup, same pattern as `consolidationTimeoutMs`. Cloud users are unaffected on the success path — the value is a ceiling, not a sleep. Documented in README.
- **Budget linked into an `AbortController`** combining the session signal with wall-clock expiry (the family's existing `AbortController` + `setTimeout` pattern). This bounds the `llmFallbackModels` chain inside one `runDirectMemoryCompletion` call, which otherwise restarts its per-model timer on every continue.
- **No more silent swallow.** When the budget runs out (or the child throws) before the flush saves, compact surfaces one `ctx.ui.notify` warning naming `flushCompactTimeoutMs`, so the remedy for slow/local models is one line in settings. Escape during compact still cancels immediately (silent). The shutdown flush keeps its own hardcoded 10s cap and stays silent.
- Non-positive `flushCompactTimeoutMs` degrades to an immediate skip: no LLM call, no doomed child spawned with a non-positive timeout.
- Transports untouched: `review-memory-ops.ts` and `pi-child-process.ts` are unchanged; both already honor `timeoutMs` + `signal`.

## Validation

- `npx tsc --noEmit`: clean.
- Targeted: `npx tsx --test tests/handlers/session-flush.test.ts tests/config.test.ts` — 84 pass, 0 fail.
- Full suite: **all 52 test files pass** on this branch.
- Pure-policy tests cover the remaining-budget math (exact values, both floor directions, tiny budgets, non-positive budgets), and the behavioral table asserts the child's exec timeout is `leftover + 5000` — never a copied budget — plus an explicit regression guard against re-introducing skip-on-abort semantics.

Fixes #225


chandra447:
Ran the full adversarial interrogation on this PR (4 independent reviewers, plus the pre-merge suite). Verdict: the one-shared-window core is sound — 4/4 reviewers verified the linked-signal swap, the leftover math, Escape semantics, config compat, and the pi.dev contract (handler shapes and `ctx.ui.notify` checked against the installed SDK typings). No criticals. The review surfaced four real gaps, all now fixed on the branch (commit `472f271`):

1. **The child-phase notify had a hole (3/4 reviewers, the sharpest find).** `pi.exec` resolves — never rejects — on timeout/kill, so a watchdog-killed child (exit 124) returned silently while a near-unreachable throw path notified. The PR's own doc table and the intent ("compact misses notify once") were wrong for the most common miss. Fixed: the resolved result is inspected (`code !== 0 || killed`) and notifies with the real outcome; the throw path now reports the actual error. The pinned test used a mock `pi.exec` that throws, which real `execCommand` never does — replaced with the resolve-`{code:124}` shape and a new test.
2. **Non-positive budget misreported (3/4).** `flushCompactTimeoutMs: 0` warned "timed out after 0ms" on every compaction — an explicit disable mislabeled as a failure. Fixed: non-positive budgets skip silently; non-finite values (e.g. `1e999`, valid JSON) fall back to the default instead of degenerating into a ~1ms silent no-op.
3. **Type-level cutout (3/4).** The required `MemoryConfig` field broke five untouched test-file literals invisibly (`tsconfig` only checks `src`). Fixed by making the field optional like its siblings, defaulting at the call site — no runtime change, no test churn.
4. **Wiring regression shipped green (2/4).** No test asserted that `runDirect` receives the *linked* signal. Added one: a stub records the signal it receives, asserts it is not the raw session signal, and resolves only when the linked signal aborts (synchronously through `linkBudget`).

Also added: CHANGELOG entries for the default raise and the new visible warning (stock-user deltas: a slow-failing direct now forfeits the child fallback by design — single window; forced-subprocess wait grows from ~35s to ~65s; both intentional per the issue), and the README ceiling wording now mentions the ~5s watchdog grace.

Full suite: 52/52 files, tsc clean.

chandra447:
Rebased onto `main` (the `session-flush.ts` conflict was against the lazy-init guard from #224; both are kept) and pushed review fixes in `58dedcf`.

- `flushCompactTimeoutMs: 0` or lower is the documented disable, so `loadConfig` no longer warns about it as a too-low timeout. Values in `(0, 60000)` still warn.
- The child miss check now reads `killed` as well as the exit code. `pi.exec` resolves the signal-kill shape as `{ code: 0, killed: true }`, which the exit-code-only check reported as a silent success.
- The budget-exhausted notice distinguishes a spent window from a leftover below the 5s spawn floor instead of calling both "timed out after N ms".
- The non-finite config test now writes raw JSON. `JSON.stringify` turns `Infinity` into `null`, so the `Number.isFinite` guard was never exercised.

Tests: `tests/config.test.ts` + `tests/handlers/session-flush.test.ts` are 90/90, full suite 55/55, `tsc --noEmit` clean, CI green.

PR #235, opened 2026-09-10
fix(review): recover thinking-channel output and stop empty-response double-spend (#197)

### LINKED ISSUE #197: Background review always fails when provider defaults to thinking: empty content, text lands in reasoning
## 现象

配置 `llmThinkingOverride` 为 `off`（官方默认）时，后台 review 依然稳定报错：

```
Memory auto-review failed in both transports. Direct: <...>. Subprocess: <...>.
```

而用 `high` 时同样失败。已排除模型链路问题（手动测试模型连通正常、鉴权正常）。

## 环境

- pi-hermes-memory 0.9.6
- 本地 vLLM 后端（DeepSeek V4 Flash），服务端启动配置 `DEFAULT_THINKING=max`
- 直接调用该 vLLM 时，响应结构为：

```
message.keys: ['role', 'content', 'refusal', 'annotations', 'audio', 'function_call', 'reasoning']
  content: None
  reasoning: "我们理解当前对话……"    # 全部输出落在 reasoning
```

## 代码路径（推测根因）

1. `buildDirectReviewCompletionOptions`（review-memory-ops.ts）：
   ```ts
   if (model.reasoning && thinking && thinking !== "off") {
     options.reasoning = thinking;
   }
   ```
   —— **`thinking = "off"` 时不传任何 reasoning/thinking 参数**，依赖服务端默认。若服务端默认开启了 thinking（如 `DEFAULT_THINKING=max`），请求仍会进入思考模式。

2. vLLM 思考模式响应：`message.content` 为 `null`，文本全部在 `message.reasoning`。

3. `responseText` 只从 `content` 数组中提取 `type: "text"` 块，**不读 `reasoning`** → 拿到的文本是空字符串。

4. `parseReviewOperations("")`：
   - `/nothing to save/i.test("")` 为 false，不会提前返回 `[]`
   - `extractJsonPayload("")` 返回 null
   - 整体返回 `null` → 被判定为 `parse_error` → direct 失败 → 回退 subprocess（同样逻辑）→ 双失败并弹出 warning。

即：**服务端默认开启 thinking 时，review 无论 `llmThinkingOverride` 设什么都会失败**——off 时不传参被服务端默认覆盖，high 时思考输出可能吃掉全部 max_tokens 导致 content 为空。

## 复现步骤

1. 本地 vLLM 服务设置 `DEFAULT_THINKING=max`（或其他非 off 默认）
2. pi-hermes-memory 配置 `llmThinkingOverride: "off"`
3. 等待一次后台 review（或手动触发）
4. 观察 "Memory auto-review failed in both transports" 提示；直接 curl 该 vLLM 可见 `message.content` 为 null、文本在 `reasoning`

## 建议修复方向

1. **thinking=off 也显式传参**：对支持 `off` 的提供商（如 vLLM `reasoning_effort=none` / `chat_template_kwargs.thinking=false`），`llmThinkingOverride: "off"` 时应显式传递给后端，而不是依赖服务端默认。
2. **responseText 兜底读取 reasoning**：`content` 为空时回退读取 `reasoning`，至少能把"服务端强制思考"识别为可诊断的情况，而不是静默 parse_error 后双失败。
3. 或者：`parseReviewOperations` 对空输入直接返回 `[]`（视为"无可保存"），避免走 parse_error 触发 subprocess 兜底重复消耗。

## 影响

- 任何把 `llmThinkingOverride` 设为 `off` 且后端默认开 thinking 的用户，后台自动学习完全失效且每次 review 双跑（direct + subprocess）浪费两次完整 LLM 调用。
- 8-23 起本环境记忆自动整理停摆，均与该路径相关。

## 环境补充

- @earendil-works/pi-coding-agent 0.84.2
- Linux / Node 24

joshua-zyy:
Fixes #197.

## Problem

When the provider defaults thinking on server-side (e.g. vLLM started with `DEFAULT_THINKING=max`), every background review fails in both transports: the answer lands entirely in the reasoning channel, `responseText()` only read `type: "text"` blocks, so the ops JSON never reached the parser → `parse_error` → the subprocess fallback burned a second full LLM call and failed the same way → "Memory auto-review failed in both transports".

## Fix

Two commits, implementing the design decisions from #197:

**`6b82f22` — recovery core**
- `responseText()` falls back to `type: "thinking"` blocks when no text block exists. Reads `.thinking` only — no `.text` alias, matching the pi-ai wire shape.
- A clean stop (`stopReason: "stop"`) with no output in either channel returns `{ ok: true, appliedCount: 0, fallbackReason: "empty_response" }` instead of a parse error — same "nothing to save" settle, distinct reason for diagnostics, no subprocess burn. Truncated (`stopReason: "length"`) responses still fall through to `parse_error` so the chain can retry.

**`dc9e80f` — hardening per the design pass**
- **Schema neutered**: the filled `add` example is gone — the old schema text itself parsed into a live `add` through `extractJsonPayload`'s first-to-last-brace slice, which is what made thinking-channel parsing unsafe at all. The shape is now shown with a deliberately unparseable placeholder, a guard comment forbids valid examples while thinking is a parse source, and the schema ends with "Put the JSON in the assistant text, not only in thinking."
- **Redacted thinking blocks skipped**: Anthropic `redacted_thinking` arrives as `{type:"thinking", redacted:true}` after normalization; a redacted-only completion settles `empty_response` instead of parse_error, so those users don't hit the subprocess double-spend either.
- **Last parseable operations object**: when the existing extraction paths (direct parse → fence → first-to-last slice) all fail, a string/escape-aware balanced-brace scan returns the last object carrying an `operations` array — CoT restates the schema before answering, so the real answer is the trailing object.
- **Timeout literals collapsed**: the three `120000` sites (review-memory-ops + both in background-review) now share `REVIEW_COMPLETION_TIMEOUT_MS`. No new config key — the double spend was the bug, not the number.

## Implementation note: last-object scan scope

The design pass scoped the last-object preference to "when the thinking channel does parse". It's implemented as the final fallback in the shared `extractJsonPayload`, so it applies to both channels: every parse that succeeds today is unchanged (the scan only runs after all existing paths fail), so it only converts former parse errors into recovered ops. Happy to tighten it to the thinking channel only if preferred.

## Tests (4 existing + 8 new)

- reasoning-only recovery, text-over-thinking precedence, clean-stop-empty → `empty_response`, length-truncation → `parse_error`
- schema echo across all four direct prompts parses to no live operations
- redacted-only → `empty_response`; redacted + clear block → clear one recovered
- CoT-then-JSON applies the trailing user add end-to-end (the exact case from the design pass)
- whitespace-only text does not mask thinking
- first-to-last slice parity (single object with prose still parses); no-ops-object input → null

## Verification

`tsc --noEmit` clean, Biome clean. Rebased onto current `main` (9885a89). Suites: review-memory-ops 47/48 (the 1 failure is a pre-existing Windows EBUSY sqlite cleanup issue, reproduced identically on clean master), background-review 34/34, session-flush 25/25, auto-consolidate 35/35, correction-detector 53/53, thinking-max 2/2.


chandra447:
Thanks for this. The recovery core is right, and five pieces from the direction I gave on #197 landed exactly as asked: the schema neuter with its guard comment, `ok: true` plus the distinct `empty_response`, the `redacted: true` skip, reading only `.thinking`, and collapsing the timeout literals into one constant. Keep all of them.

Going back over the branch, I found a few things I missed when I wrote that direction, plus one refinement from it that landed somewhere it does not run. Three changes are needed before merge.

## 1. Extraction accepts any JSON object, so a non-ops snippet recreates the double-spend

Every step before the new scan accepts a candidate without checking that it carries `operations`: whole text at `:324`, the fence at `:329-332`, the first-to-last slice at `:339-344`. Only `lastParseableOperationsObject` (`:299`) validates. When one of those steps claims a payload that has no `operations` array, `parseReviewOperations` returns `null` (`:373-375`), and the result is `parse_error` into the fallback chain and the subprocess.

Reproduced against the head, with a valid trailing answer present the whole time:

```
'```json\n{"note":"no ops here"}\n```\nFinal:\n{"operations":[{"action":"add","target":"user","content":"real answer"}]}'
  -> null
```

Any well-formed JSON object in the reasoning, a log line, a counterexample, an echo of another prompt fragment, is enough. The last-object scan never gets to run. The fix is small and strictly dominant, because those candidates resolve to `null` today anyway:

```ts
const asObject = (value: unknown): JsonObjectPayload | null =>
  isJsonObjectPayload(value) && Array.isArray((value as { operations?: unknown }).operations)
    ? (value as JsonObjectPayload)
    : null;
```

## 2. The thinking-channel preference needs its own extraction path

The direction I gave scoped the last-object preference to the thinking channel. It is currently the final fallback of the shared cascade, which is why both failure modes below are live.

- A fenced draft earlier in the reasoning wins before the scan runs, so a chain-of-thought draft becomes a live mutation.
- `topLevelObjectSpans` (`:265`) only records spans that return to depth 0, so one unbalanced `{` in prose holds depth for the rest of the channel and the real answer never becomes a top-level span.

Please give thinking-sourced text its own path: validate candidates on the `operations` array, scan from the end, and decide trailing-ness from the winner's span. Leave the text-channel order untouched so today's parse successes do not move.

## 3. `empty_response` should also cover thinking-only output that does not parse

The gate at `:704` requires both channels to be empty, so the common thinking-max shape still walks into `parse_error` and then the subprocess: thinking full of brace-heavy prose, no trailing ops object, text empty. The subprocess runs the same model against the same server-side thinking default, so that second call is provably doomed. That is the reporter's original "failed in both transports" coming back.

Treat thinking-sourced output with `stopReason: "stop"` that parses to nothing as `empty_response` rather than `parse_error`. Truncation stays `parse_error`, the same line the current code draws for empty content at `:702-703`.

## 4. `empty_response` must walk `llmFallbackModels` before settling

Today it returns out of the whole loop at `:706`, so a silent primary model ends the review and the configured chain never runs. `parse_error` does continue to the next model, so the two outcomes currently disagree about whether a broken model is worth retrying.

Replace the return with the existing `lastResult` plus `clearTimeout` plus `continue` pattern used at `:690` and `:711`, and keep the `{ ok: true, appliedCount: 0, fallbackReason: "empty_response" }` contract after the last model. Never the subprocess on this path.

## 5. Consolidation must be terminal on `empty_response`

`triggerConsolidation` gates success on `directResult.ok && directResult.appliedCount > 0` (`src/handlers/auto-consolidate.ts:221`), so `empty_response` (ok, zero applications) fails the gate and falls through to the consolidation lock and a `pi -p` child at `:232` on the same broken model. This is the one path where the double-spend survives the PR.

Consolidation's success criterion does not change. It must still shrink. The change is only that an empty completion is terminal: return `{ consolidated: false, error: "<provider> returned an empty completion; no consolidation attempted" }` instead of spawning the child.

## Behaviour note for the trailing decision

With a thinking-sourced payload, the trailing candidate is the answer, so all three actions are allowed. A parsed candidate that is not trailing is draft-grade, so apply `add` only from it and drop `replace`/`remove`. When the trailing candidate is an empty `{"operations":[]}`, settle as empty rather than reaching back to an earlier non-empty candidate. That loses a save in the case where the model answers and then quotes the schema's own empty example, and it is deliberate: applying an earlier candidate can execute a draft the model rejected, and a missed save is recoverable while a wrong deletion is not.

## Tests to add

- Fenced non-ops object before a valid trailing answer parses the answer.
- Fenced draft before a trailing answer picks the trailing answer, and the draft's `remove` is not applied.
- Non-trailing parsed payload applies `add` only.
- Thinking-only unparseable output with a clean stop settles `empty_response` and never spawns the subprocess.
- A silent first model walks to a healthy second model and applies its operations.
- Consolidation returns terminal on `empty_response` without acquiring the lock or spawning a child.
- Existing four direct prompts still parse to no live operations.

joshua-zyy:
All five changes are in, pushed as `e773bec` (rebased onto `c4a3289` first). Every item reproduced locally before implementing — including your fenced non-ops repro, the fenced remove-draft winning via the cascade, and the unbalanced-brace case — and all three now resolve to the trailing answer.

**1. Ops validation with fall-through.** One note beyond the snippet: `return asObject(...)` in the early paths exits the whole extraction when the candidate is declined, so the fence branch would have returned null without the scan ever running. Declined candidates now fall through to the later paths (`if (parsed) return parsed`), which is what actually lets the scan recover the trailing answer.

**2. Thinking path.** Thinking-sourced text skips the cascade entirely: ops-validated candidates scanned from the end, winner's trailing-ness decided from its span, and an all-depth balanced-span scanner so an unbalanced `{` in prose cannot hide a later object. The text channel keeps the cascade untouched. Trust rules as specified: trailing → all actions; non-trailing → adds only; trailing empty → settle empty, no reach-back.

One definition to confirm: I read "trailing" as *only whitespace after the winner's closing brace*. That means a thinking channel ending with a fenced answer (`...```json\n{ops}\n```) is non-trailing — the closing fence follows the object — so it gets draft-grade add-only treatment. Strict felt like the conservative reading of "a missed save is recoverable, a wrong deletion is not", but if you'd rather count a trailing fence closer as still trailing, it's a two-line change.

**3–4. empty_response.** Covers thinking-only output that parses to nothing on a clean stop (truncation still parse_error), and walks llmFallbackModels with the same lastResult/clearTimeout/continue pattern as parse_error — silent primary, healthy fallback, operations applied; both silent → empty_response after the last model.

**5. Consolidation terminal.** Returns `{ consolidated: false, error: "<provider/model> returned an empty completion; no consolidation attempted" }` before the lock, no child.

Tests: your six new cases, plus three from the failure modes and behaviour notes (unbalanced-brace recovery, trailing-empty no reach-back, truncation-stays-parse_error). Verification: tsc/Biome clean; review-memory-ops 55/56 (the 1 failure is the pre-existing Windows EBUSY sqlite cleanup issue, reproduced identically on clean master), auto-consolidate 36/36, background-review 34/34, session-flush 54/54, correction-detector 53/53, thinking-max 2/2.

chandra447:
Confirmed on all of it. And yes, your reading of "trailing" is the one I want: only whitespace after the winner's closing brace counts as trailing. A fenced answer with the closing fence after it is non-trailing, so it gets draft-grade add-only treatment. That is deliberately strict. The trailing position is the only structural evidence that the model finished reasoning and committed to the object. Anything that follows it, a fence or prose or another object, says the object was produced mid-stream and possibly superseded. A missed save is recoverable; a wrong deletion is not. Keep the strict reading exactly as you implemented it.

I also re-verified the head (`e773bec`) against the double-spend cases rather than just reading the patch:

- the fenced non-ops snippet before a trailing answer now falls through and the trailing answer wins;
- an early draft containing a `remove` no longer wins over the trailing answer;
- an unbalanced `{` in prose no longer hides a later object;
- a trailing empty candidate settles empty with no reach-back;
- the text channel's cascade is untouched, so today's parse successes do not move.

Your fall-through note is correct, by the way. `return asObject(...)` in the early paths would have exited the whole extraction when the candidate was declined, and the fall-through is what actually lets the last-object scan recover the answer. That was the second half of the bug, and the repro alone would have hidden it.

Tests: 56 pass in the review-ops suite, `tsc --noEmit` clean. Nothing else from my side. This is ready to merge.

PR #236, opened 2026-09-11
fix(consolidation): split oversized subprocess consolidation into bounded, resumable rounds

dawidmachon:
## Problem

Auto-consolidation on the subprocess path sends the whole store to one `pi -p`
child under a single `consolidationTimeoutMs` (default 180s). At cap scale the
multi-turn merge exceeds it and gets killed:

    Consolidation subprocess was terminated (likely timeout). Timeout: 180000ms.

Observed in production with memory at 4,922/5,000 chars. Store stays over
capacity and subsequent adds keep failing.

## Fix

Above `consolidationChunkChars` (new, default 4000) the subprocess path runs
bounded rounds that share one overall time budget: a slice of entries per
child run, each under `min(consolidationTimeoutMs, remaining)`, reloading from
disk between rounds.

- Stores ≤ threshold: byte-identical single-shot behavior.
- Direct in-process transport (`/memory-consolidate`): untouched (all-or-nothing).
- Rounds stop at the target's **capacity goal** (`charLimit(target)`, failure
  tier = 2× memory limit) — not at the prompt budget, so a 2×-tier store is
  not over-merged down to one slice size.
- The rounds share `consolidationTimeoutMs` as a total budget: a chunked
  trigger never blocks longer than the single call it replaced. Starvation
  reports `partial` and the next trigger resumes from disk state (no cursor
  needed — partial progress is already on disk).
- A slice that shrinks nothing is skipped; the walk covers later slices
  instead of blocking on an unshrinkable head. When the whole walk yields
  nothing, the run reports the honest dead end.
- Round failures after earlier progress report `consolidated: true` +
  `partial: true` + reason; zero completed rounds keep the old failure shape.
  All consumers surface it: warning path, store add-retry error,
  `/memory-consolidate` (`⚠️ partial: …`).
- Chunked rounds get a prompt scope guard; disappearances outside a round's
  slice (rogue child deletions, legitimate cross-slice dedup, or concurrent
  writers — indistinguishable to the parent) are detected and **reported,
  never resurrected**: re-adding through the cap-enforcing add path cannot
  converge in legacy-inject mode and resurrection fights legitimate dedup
  across triggers (#226 class). Recovery snapshots remain the repair path.
- An over-chunk store already within its capacity goal is a clean no-op
  (`consolidated: true, rounds: 0`).
- When a target's capacity goal sits below the chunk size (small caps), the
  final round — once the remaining store fits one prompt — runs unscoped with
  the whole remaining store for full context.

## Config (backward-compatible)

| key | default | meaning |
|---|---|---|
| `consolidationChunkChars` | 4000 | split subprocess consolidation above this; min 500 |

`consolidationTimeoutMs` (existing) now also bounds the total chunked trigger
time. No other new knobs.

## Proof

- New tests: multi-round goal-driven completion, resume-after-kill (child
  rewrites the store **file** directly; parent reload observes a separate
  writer), no-progress slice walk + dead-end reporting, per-round shared
  budget, out-of-scope detection without resurrection, legacy-inject
  cap-enforced run (store genuinely over limit), prompt scoping, config
  parsing, `/memory-consolidate` partial display. Full suite green.
- Real end-to-end runs (real child pi, real model): 2×-cap store → 2
  goal-driven rounds, `partial: false`, zero entries vanished outside their
  slices; live store under its goal → clean 18 ms no-op.


chandra447 (changes requested):
Reviewed the diff at `31d21f4` against the pi.dev extension docs and a minimality lens, with four independent adversarial passes.

The mechanism itself is sound. Below the threshold the single-shot path is byte-identical, slices are whole entries, resume needs no cursor because partial progress is already on disk, and the new tests cover multi-round, resume-after-kill, no-progress, and the per-round timeout. The problem is what the loop reports and where it stops.

## 1. A run that damaged the store still reports success

The chunked path returns `{ consolidated: true, rounds, error? }`, and every consumer branches on `consolidated` alone:

- `src/index.ts:312` warns via `shouldWarnAutoConsolidationFailure(config.autoConsolidationWarnOnFailure, result.consolidated)`, and `src/auto-consolidation-warning.ts:5` is literally `return !consolidated && warnOnFailure;`
- `src/store/memory-store.ts:317` reads `error` only inside the `!consolidated` branch
- `/memory-consolidate` pushes a literal `✅ consolidated` in its `result.consolidated` branch

So a round killed after a successful first round reads as success, and so does the out-of-scope wipe the last commit was written for. Before this change, the same kill produced `❌ Consolidation subprocess was terminated (likely timeout)`. The field is set, but nothing in the product reads it.

Put the outcome in the result shape instead of trusting consumers to notice a field: `consolidated: false` when no round completed, plus `partial: true` when rounds completed but the run ended with an error. Then the three call sites warn on `partial`, print `result.error` beside the `✅`, and log the error on success-with-error.

## 2. The loop exits on the prompt budget, not on the capacity goal

The loop breaks when `promptEntries.join(ENTRY_DELIMITER).length <= chunkChars` (4000). The store's actual goal is `charLimit(target)`, which for the failure target is `memoryCharLimit * 2` — 10000 by default (`src/store/memory-store.ts:120`). A failure store that overflowed at ~10001 chars keeps merging until it is under 4000, and every one of those rounds is told to be aggressive about merging. Slice size and capacity are different concepts; breaking on the store fitting its target again would also remove most of the added latency in the common case.

## 3. No overall time budget

The overflow path calls `triggerConsolidation` without `directCtx`, so it always takes the subprocess path, inside a blocking `memory_add`. Worst case becomes `maxRounds * consolidationTimeoutMs` = 4 x 180s = 12 minutes, against 180s before this PR.

## 4. Scope is enforced by prompt text against a child that can see the whole store

The parent's own real-run testing found a round wiping 11 out-of-scope entries. The defenses are a prompt footnote plus a post-hoc vanished-entry note, and per item 1 that note never reaches a user. Structural options worth considering: run the child against a slice-scoped store, or restore the vanished entries from the existing `.recovery-*` snapshot when they disappear.

## Smaller, non-blocking

- The mock child mutates the same in-memory store the parent holds, so the reload-from-disk path is asserted through in-process state rather than files a child actually wrote.
- `consolidationChunkChars` / `consolidationMaxRounds` defaults live in the handler instead of `DEFAULT_CONFIG`, and the range check is duplicated between `src/config.ts` and the `*For` helpers.
- `ConsolidationResult.rounds` has no consumer, and `consolidationMaxRounds` may not earn its place as a knob.
- No CHANGELOG entry.
- CI has not run on this branch ("no checks reported"), so tests and typecheck are unverified in CI.

## One open question

What should happen when the head slice cannot shrink because every entry is a distinct fact? Today the no-progress guard breaks the loop and returns `consolidated: true` with one round while the store stays over capacity.

dawidmachon:
Hello, just letting you know, I really appreciate your review and extension. I checked in past many extensions and ideas, but non of them following into space how this blog describes: www.pinglin.tw/blog/the-shapes-of-agent-memory/ 

I picked your as most complete approach + planing to focus on some, small, modular PR which will aim to close gap into modern approach for token efficient memory. 

Also leting you know, i reading by myself all what you have, rethinking and feeding agent as needed, to not bloat you with too much slop ;D

Of course, give me hints and I will try to follow your guidance. I will push my ideas as PR one at the time, when test a bit myself, so fell free to diccuss over them and think if that suits for your idea of your repository ;)

Big thanks for you to work over this!

p.s.
I will adress your notes of course later today/tommorow ;)

dawidmachon:
Thanks for the review — all four points check out against the code, and I push `953d5c8` addressing each. Per-point:

**1. Success reporting on damaged runs — agreed, fixed.** Verified all three consumers branch on `consolidated` alone. The chunked path now returns `partial: true` whenever rounds completed but the run ended with a failure or the store is still over its capacity goal; `consolidated: false` only when nothing completed. `shouldWarnAutoConsolidationFailure` takes `partial`, the store's add-retry appends the reason when the retried add still fails, and `/memory-consolidate` prints `⚠️ partial: <reason>` with the round count.

**2. Prompt budget vs capacity goal — agreed, and it bit me in live testing too.** The loop now stops at `charLimit(target)` (exposed as `MemoryStore.capacityGoal()`), not at `chunkChars`. Verified live: a 4,160-char memory store (over the 4,000 chunk, under the 5,000 goal) now no-ops in 18 ms instead of spawning rounds. Two derived behaviors: when the *remaining* store fits one prompt, that decisive round runs **unscoped** (whole remaining store, full context), and the failure tier stops at 10,000 instead of being over-merged to 4,000.

**3. Time budget — agreed, fixed without a new knob.** The rounds share one overall budget equal to `consolidationTimeoutMs`, so a chunked trigger never blocks longer than the single call it replaced. Each round gets `min(consolidationTimeoutMs, remaining)`; starvation reports partial and the next trigger resumes from disk state.

**4. Prompt-only scoping — agreed it was weak.** Middle ground implemented now: the parent restores out-of-scope vanished entries (re-adds the pre-round text) and reports restored/ unrestored counts in `error`. The fully structural option (slice-scoped child store with merge-back) I'd propose as a separate PR if you want it — it touches the store bootstrap path and deserves its own review.

**Open question (unshrinkable head slice):** the loop no longer breaks — it *rotates* to the next slice, so an unshrinkable head no longer blocks the tail. When the whole store has been walked without progress, it returns `partial: true` with "no slice could be shrunk toward the capacity goal; entries may be distinct facts worth keeping". Honest dead-end reporting rather than a fake ✅.

**Smaller items:**
- Mock children in the new tests now rewrite the store **file** directly; the parent's reload observes what a separate writer wrote.
- `consolidationMaxRounds` dropped (the budget bounds time; an internal `MAX_CONSOLIDATION_ROUNDS` guards runaway loops). README + example updated.
- `rounds` now feeds the `/memory-consolidate` output (`✅ consolidated (2 rounds)`).
- CHANGELOG entry added under Unreleased.
- Range validation stays in `loadConfig`; the handler-side fallback covers direct API callers with partial config objects — can unify if you prefer.
- CI: no checks were reported on the fork branch — if workflows need approval for first-time contributors, happy to re-run anything you want locally and paste output.

Verified after the rework: `tsc --noEmit` clean; auto-consolidate/config/index/memory-store/flow suites pass per-file (the repo's own run-all convention); two fresh real end-to-end runs (real child pi, real model) — 2×-cap store completes in 2 goal-driven rounds (`partial: false`), and the live over-budget-but-under-goal store correctly no-ops.

chandra447 (approved):
This addresses all four items, and I verified each against `953d5c8` rather than taking the write-up at face value.

1. **Partial reporting.** `warnOnFailure && (!consolidated || partial)` now warns on partial progress, `consolidated: false` happens only when zero rounds completed, and `/memory-consolidate` surfaces `⚠️ partial: <reason>` with the round count. The success-after-damage hole from my review is closed.
2. **Goal-driven rounds.** The loop gates on `capacityGoal(target)` instead of the prompt budget, the decisive round runs unscoped when the remaining store fits one prompt, and the failure tier stops over-merging at the goal. The 4,160-char no-op you measured is the behavior I asked for.
3. **Time budget.** One shared deadline equal to `consolidationTimeoutMs`, per-round `min(consolidationTimeoutMs, remaining)`, and `consolidationMaxRounds` is gone from config and types, replaced by an internal round cap. Right call on minimality; that config was a knob users would have needed to understand to avoid footgunning themselves.
4. **Scope-exit restoration.** Re-adding pre-round text for out-of-scope vanished entries, with restored and unrestored counts in `error`, is the right middle ground. Agreed the slice-scoped child store with merge-back belongs in its own PR; it touches the store bootstrap path and deserves its own review.

On your open question about the unshrinkable store: rotating to the next slice and reporting `partial: true` with the "no slice could be shrunk" message is exactly right. Honest dead-end reporting beats a fake ✅, and "entries may be distinct facts worth keeping" is the correct hint for the user.

Verification on my side: `tsc --noEmit` clean on the head, 49 tests in the auto-consolidate suite and 135 across the config/index/memory-store suites all pass, and the restored/unrestored accounting path checks out.

One thing before merge: the branch now conflicts with `main` in two files, and both are keep-both.

- `src/constants.ts`: your consolidation constants with their doc comments and `main`'s flush-compact constants both stay.
- `src/config.ts`: keep both imports, `CONSOLIDATION_CHUNK_CHARS_MIN` and `DEFAULT_FLUSH_COMPACT_TIMEOUT_MS`.

Rebase onto `c4a3289` and this is ready to merge.

chandra447 (changes requested):
I have to walk back my approval. Going through the branch again with the intent to merge, I verified that the out-of-scope restore, the piece this rework added as the backstop for scoped rounds, is structurally broken in the mode where chunked consolidation actually auto-runs, and the new tests cannot have caught it. The design survives; the restore path does not.

## 1. The restore cannot repair in the mode it exists for

The restore loop re-adds vanished out-of-scope entries through the public `store.add` (`auto-consolidate.ts:412`). Three problems ride on that one call.

`add` is the cap-enforcing path. Auto-consolidation triggers precisely because a write hit the cap, and `capEnforced` is true for every mode except `policy-only`. So in the only mode where this loop auto-runs, the store is over its limit by definition, `newTotal > limit` rejects every re-add, and the entries the restore exists to protect are lost anyway.

The rejected re-add does not fail quietly either. It re-enters `addWithConsolidation`, which on an overflow failure with the `auto-consolidate` strategy invokes the full consolidation trigger. The parent is holding the consolidation lease at that moment, so the nested trigger stalls on the lock wait and comes back deferred. Each restore attempt burns that wait inside the shared budget, and the failure surfaces as "another session is consolidating this target right now", which is not what is happening.

And even where an add would be accepted, the re-add loses the `<!-- created=… -->` provenance comment: the pre-round text already carries one, so it gets re-encoded with fresh metadata or double metadata, depending on which layer wins.

What the restore needs is an internal add that bypasses cap enforcement and the consolidation trigger, and preserves the original metadata. Running the repair through the same path the repair is defending against cannot converge.

## 2. The restore cannot tell a rogue deletion from a legitimate one

The child sees the real store, so an out-of-scope deletion can be a correct cross-slice dedup. Another session can delete an entry while the trigger runs. The restore treats every disappearance outside the slice as damage and re-adds all of it, so correct dedups come back as duplicates and concurrent deletions get resurrected. A deterministic child then produces the same rounds next trigger, and the store ping-pongs instead of converging. This is also a new route for the #226 resurrection class.

## 3. The new tests cannot have tested the chunked loop

The tests import `DEFAULT_CONSOLIDATION_MAX_ROUNDS` from `constants.js`, and the head exports `MAX_CONSOLIDATION_ROUNDS`. The missing named import resolves to `undefined` at runtime instead of failing, so the suite stays green while every test that touches the round cap is vacuous. The mock children also run in `policy-only` and rewrite the store file directly, bypassing the cap-enforced mutation machinery, which is exactly where item 1 lives. Tests that load a real store in `legacy-inject`, hold the lease, and exercise the cap-enforced trigger path are needed before this can merge.

## 4. Smaller but real

- `/memory-consolidate` now reports ❌ for a healthy store sitting between the chunk size and its goal ("already within its capacity goal") instead of a clean no-op.
- `offset = 0` after any progress means a large store re-works the head slice on every trigger while the tail never gets reached.
- The concurrent-writer window between rounds: entries another session legitimately deleted are misattributed to the child and "restored".
- `failureMessage` is a single string that later rounds overwrite, dropping accumulated notes.
- Several loop comments still describe per-round timeouts; the budget is shared now.

## What stands

The chunked round loop, the shared time budget, the goal-driven stop, and the partial reporting are sound and were the right fixes for the four items from my review. None of this needs a redesign. Items 1 and 3 are the merge blockers; item 2 needs at least the attribution fix, or the structural fix pulled into this PR.

dawidmachon:
Walk-back accepted — every point verified against the code, and you're right that the new tests could not have caught item 1: they ran policy-only stores and the mock children rewrote files directly, so the cap-enforced path the restore depended on was never exercised. Pushed `c80087e`:

**1. Restore removed — report-only now.** You called the mechanism: `store.add` is the cap-enforcing path, legacy-inject is over its limit by definition when the loop auto-runs, so every re-add was rejected, each rejection re-entered `addWithConsolidation` and stalled on the lease the parent already holds, and provenance was reset on the re-encode. All three die with the restore. Out-of-scope disappearances are now *reported* — note names the round and count and says "by the child or a concurrent writer", because the parent cannot attribute them. Detection without mutation can't ping-pong, and recovery snapshots remain the repair path for real damage. If attribution ever becomes possible (child reports its exact deletes back through a structured channel), that plus the slice-scoped store belong in the follow-up PR.

**2. Resurrection/attribution — resolved by the same removal.** No parent-side re-adding means no fight with legitimate cross-slice dedup, no resurrected concurrent deletions, no new route into the #226 class.

**3. Vacuous tests — root cause found and fixed.** The phantom `DEFAULT_CONSOLIDATION_MAX_ROUNDS` import was left over from the rename; `tsconfig.json` includes only `src/**`, so `tsc --noEmit` never saw the test tree and the missing named export resolved to `undefined` at runtime. Import removed (grep across `src/` + `tests/` is clean). Would you take a `tsconfig` include of `tests/` as a follow-up PR? It would have caught this mechanically, but I didn't want to widen typechecking scope inside this PR.

**New test for the mode that matters:** `chunked consolidation in legacy-inject mode` — seeds a directory past a 300-char legacy cap via a policy-only instance, reopens it legacy-inject over the same directory (store genuinely over limit), runs the chunked loop through the file-writing child, and asserts: rounds run, no cap rejection, no "another session is consolidating" stall, deletions stay reported-only.

**4. Smaller items, all in:**
- Over-chunk-but-under-goal is a clean no-op: `{ consolidated: true, rounds: 0 }` — `/memory-consolidate` shows `✅ consolidated (0 rounds)`, not ❌.
- The slice walk no longer resets to the top on progress; it advances (and wraps, with a clamp after removals), so a head that always yields a little progress cannot starve the tail.
- With the restore gone, the concurrent-writer window is down to a report note; nothing is misattributed *and mutated*.
- `failureMessage` replaced by an accumulated notes array — later rounds append, nothing overwrites.
- Stale per-round-timeout comments rewritten to the shared budget; CHANGELOG reworded (restored → reported).

**Verified after the rework:** `tsc --noEmit` clean; auto-consolidate (51), config, index, memory-store (92), flow suites pass per-file; fresh real end-to-end run (real child pi, real model): 2×-cap store → 2 goal-driven rounds, `partial: false`, zero entries vanished outside their slices.

**One more adversarial pass over my own rework** caught and fixed: the slice-walk offset could point past the shrunken store after a round's removals, producing an empty-slice round — now clamped. Also added a test for the `/memory-consolidate` partial display line, and documented precisely where the unscoped decisive round applies (only when a target's capacity goal sits below the chunk size — with the default 5000/4000 the goal exit comes first, so default-config rounds are all scoped slices and merge quality is carried by convergence across triggers).

**Known trade-offs, stated before you find them:** (a) the budget is enforced between rounds — `retryWithoutOverrides` can extend the last started round past it (bounded by one extra round timeout, and only on override-resolution failures); (b) the walk offset over a mutated array is approximate, so entries can be re-processed or skipped within a single trigger — convergence comes from the goal exit plus repeated triggers; (c) out-of-scope detection is O(n·m) over entry lists — negligible at char-bounded store sizes.

Rebased onto `c4a3289` (kept both sides in `constants.ts` and `config.ts` as you specified).

PR #238, opened 2026-09-12
fix: cap session search snippets instead of returning full message bodies

miayanlarose-star:
## What this changes

**Stated plainly up front:** on current `main`, the context flood from `session_search` is already bounded at the tool level by #100 (`snippetChars` 100–4000 per result, hard 50 KiB output ceiling, `truncatedCount`/`outputTruncated` accounting). This PR does **not** change that bounding. Its value on main is:

1. **Snippet quality**: the tool previously truncated oversized snippets with a blunt head-slice, so a 6 MB legacy row showed its preamble, not the region the query matched. Truncation now opens a **window around the first hit of the query's effective term set** — the same quoted-phrase-aware, #184 stop-word-filtered terms the FTS5 match itself used — so the visible window contains what actually matched.
2. **Defense in depth** for the older generation path: an all-stop-word query, or a row matched by the LIKE fallback on raw terms, has no surviving anchor term; the window then falls back to the head (the previous behavior) — a **documented fallback, not a promise kept**.

The 11.7 MB full-body snippet detonation applies to **released versions whose tool predates #100** (e.g. 0.7.23, where `session_search` returned entire message bodies as rendered snippets); on this head it was already fixed by #100.

## Implementation

- **One truncation owner — the tool.** The first iteration of this PR put a 900-char cap in the store, which broke the #100 output contract (the tool's knob could no longer do anything, the two truncation notes contradicted each other, and the tool suite failed 3 of 8). Per review, the window is now the implementation of the tool's *own* default truncation: `truncateLegacySnippet` takes the anchor terms, `snippetChars` stays the ceiling the window respects, and the note format, 50 KiB aggregate cap, and `truncatedCount` accounting are untouched. The store's shape is unchanged apart from the item below.
- **Anchor = the matcher's term set.** `collectNaturalLanguageTerms` (exported from `fts-query.ts`) is the same collector `normalizeFts5Query` builds the FTS5 query from: quoted phrases preserved, connectors and #184 stop words dropped. Anchoring on raw tokens would let a stop word near the head of a long body pin the window at the start — exactly what windowing exists to avoid.
- **Cost on pathological rows:** the body is lowercased **once per result**, not once per term; each term is a single native `indexOf` over the shared lowercase copy.
- **No double full-body materialization:** the store previously selected `m.content` *and* `m.content as snippet`, materializing the full body twice per row (identical values). It now selects `m.content` once and maps both fields from it. `content` remains in `SessionSearchResult` (store tests read it); the tool reads only `snippet`.
- **Surrogate safety:** window edges are UTF-16 code units; if an edge lands between the halves of a surrogate pair it is nudged by one unit, so a sliced emoji cannot render corrupted.

## Tests

The #100 tool suite (3 of 8 were failing on the previous iteration) passes **unchanged** — the contract it pins (≤50 KiB, note format, `6000007 chars total`, `truncatedCount`, no output echo in details) is preserved by construction. Two new tests pin the window deliberately:

- **windows an oversized legacy snippet around the first effective term hit, not the head** — a `the …` head, the needle ~4 KB deep; the window must contain the needle region, start with the `…` elision marker, and not show the head.
- **keeps surrogate pairs whole at window edges** — a needle placed so the lead-adjusted window start and the odd-`snippetChars` window end both land on the LOW half of an emoji pair; the sliced text must contain only whole code points.

Full suite: 55 test files, all green; `tsc --noEmit` clean.


chandra447 (changes requested):
The problem is real and the instinct is right — the 11.7 MB legacy row is exactly the kind of thing that detonates an agent context, and a window around the first hit is a better snippet than a blunt head-slice. But the change as written sits at the wrong layer, breaks the existing output contract, and I verified the breakage against the head before writing this.

## Blocker: the tool's own integration tests fail at this head

I ran the tool suite against the branch: 3 of 8 tests in `tests/tools/session-search-tool.test.ts` fail (`bounds oversized legacy results and reports truncation without duplicating output in details`, `offers a bounded snippetChars override for legacy searches`, `enforces a hard 50 KiB ceiling across many large legacy results`). The PR's testing section reports the store suite green, but the tool suite owns the output contract from #100 and this diff changes what that contract observes without touching it. That suite has to be green before merge, and its assertions need to be updated to the new shape deliberately, not discovered here.

## The layering conflicts with the #100 contract

On `main`, the session-search tool already owns truncation: `truncateLegacySnippet` slices per result to the user-facing `snippetChars` parameter (100 to 4000, default 1200), `capLegacyOutput` enforces a hard 50 KiB ceiling, and the result reports `truncatedCount` / `outputTruncated`. The store-level 900-char cap now truncates first, so the tool's knob can no longer produce more than about 940 chars for any long message, the two truncation notes contradict each other, and the tool's accounting no longer reflects the real lengths.

My recommendation: make the windowed cap the implementation of the tool's own default, not a second layer underneath it. One truncation owner, one note format, `snippetChars` stays the ceiling the window respects, and the #100 tests get updated to the better window. If you prefer keeping the cap at the store layer, then the tool's accounting has to be re-derived from the capped snippets and the interplay with `snippetChars` documented, which is more code for the same outcome.

Worth stating plainly in the PR description too: on `main` the context flood is already bounded at the tool (that is what #100 did), so this PR's on-main value is snippet quality plus defense in depth; the 11.7 MB detonation applies to released versions whose tool predates #100. Both are legitimate reasons to land it, but the description currently reads as if `main` is unprotected.

## The window anchor is not the matcher that produced the result

Windowing searches for `collectLikeTerms(query)` hits, but FTS-matched results were matched by the FTS tokenizer with #184's stop-word filtering, and `collectLikeTerms` keeps stop words and sub-3-character tokens. For a natural-language query — the common case against long messages — the first "hit" is usually a stop word near the start of the blob, so the window degrades to the first 900 chars, which is precisely the long-message scenario this PR exists to fix. Use the same effective term set the FTS path matched on (stop words filtered the same way), and treat no-hit as a documented fallback rather than a promise kept.

## Cost on the pathological rows

`buildSnippet` lowercases the entire content once per term, per result. On the motivating rows that is the difference between an O(1) slice and a full-body scan multiplied by the term count, times the result count. Precompute one lowercase copy per result, or do a single case-insensitive scan.

`content` still carries the full body in every result object and no caller reads it. Fetching it costs a second full-body materialization per row (on legacy indexes, megabytes), and carrying it re-arms the flood for the next caller that renders it. Either drop it from the capped results or gate it behind an option for callers that actually need it.

## Smaller items

- `slice(start, end)` and `.length` are UTF-16 code units; a surrogate pair at a window edge renders corrupted, and the note says "chars".
- 900 and 200 are a third cap family with no shared constant, in a different home from the repo's other caps (`getMessageText`'s 500, the tool's `snippetChars`, `CONSOLIDATION_CHUNK_CHARS_MIN`).
- The changelog says "capped at 900 characters", but the appended note pushes the total past 900; and the note's wording drifts from the indexer's existing truncation note.

miayanlarose-star:
Thanks for the careful review — every point was addressable, and the branch is rewritten as a single clean commit (`61586b3`, force-pushed) rather than a pile of fixes. Point by point:

- **Tool suite failures / layering**: the store-level 900-char cap is gone entirely. The window is now the implementation of the tool's own `truncateLegacySnippet`: `snippetChars` stays the per-result ceiling the window respects, the 50 KiB aggregate cap, note format, and `truncatedCount` accounting are untouched, and the 3 previously-failing #100 tests pass **unchanged**. One truncation owner.
- **Window anchor vs. matcher**: `collectNaturalLanguageTerms` is now exported from `fts-query.ts` and is the anchor source — the same quoted-phrase-aware, #184 stop-word-filtered term set `normalizeFts5Query` builds the FTS5 query from. No-hit is an explicit, documented head fallback (in the JSDoc and the PR description).
- **Cost on pathological rows**: one lowercase pass per result; each term is a single native `indexOf` over the shared copy.
- **`content` duplication**: the store no longer selects `m.content` twice — it fetches the column once and maps both `content` and `snippet` from it. `content` stays in the interface because the store tests read it, but nothing renders it.
- **Surrogates**: window edges are nudged off pair splits; there's a deterministic test that places both the start (lead-adjusted) and end (odd `snippetChars`) on a LOW half of an emoji pair and asserts only whole code points survive.
- **Constants home / changelog wording**: the lead-context constant lives beside the tool's other legacy truncation constants with a pointer to the indexer's 100 KB cap and `getMessageText`'s 500; the changelog entry no longer claims a 900-char cap and describes the window, the fallback, and that #100 already bounds the flood on main.

Also stated plainly in the rewritten PR description: this PR's on-main value is snippet quality plus defense in depth; the 11.7 MB full-body detonation applies to releases whose tool predates #100.

New tests pin the window (stop-word-anchored, not head; surrogate-safe edges) alongside the unchanged #100 suite — 55 files green, `tsc` clean.

PR #242, opened 2026-09-14
fix: honor llmFallbackModels on consolidation subprocess path

chandra447:
## Summary
Re-opens the consolidation subprocess fallback work from closed #241 (`podledges/fm/pi-hermes-memory-consolidation-fallback` @ `594cbf9`).

After #215, direct in-process review/flush/correction honors `llmFallbackModels`, but automatic over-capacity consolidation still used a `pi -p` subprocess path that did not carry or apply them. Live `/memory-consolidate` on v0.9.9 could die on Codex overload even with other providers configured.

This branch wires `llmFallbackModels` through `triggerConsolidation` → `execChildPrompt` (still `pi.exec`), retries recognized provider/transport failures, reloads disk before another paid attempt and skips if memory already shrank, and stops the chain on abort. `consolidationTimeoutMs` is per model attempt.

## Before merge (one-line tighten)
Drop bare HTTP `404` from `RETRYABLE_PROVIDER_FAILURE` in `src/handlers/pi-child-process.ts` (model/provider not-found stays on `retryWithoutOverrides`). Patch is ready locally if you want a follow-up commit.

## Credit
Prior art and implementation: @podledges in #241 (closed unmerged).

## Test plan
- [ ] CI: Lint, Type Check, Tests, Min Supported SDK
- [ ] Existing auto-consolidate fallback tests pass
- [ ] Optional: force primary overload with `llmFallbackModels` set; confirm next model is tried and abort stops the chain


src/handlers/pi-child-process.ts
```diff
@@ -65,6 +71,26 @@ interface ResolveChildPiInvocationOptions {
...
+const RETRYABLE_PROVIDER_FAILURE = new RegExp([
+  String.raw`\boverloaded\b`,
+  String.raw`\bover[ -]?capacity\b`,
+  String.raw`\brate[ -]?limit(?:ed| exceeded)?\b`,
+  String.raw`\btoo many requests\b`,
+  String.raw`\btemporar(?:y|ily) unavailable\b`,
+  String.raw`\bservice unavailable\b`,
+  String.raw`\bbad gateway\b`,
+  String.raw`\bgateway timeout\b`,
+  String.raw`\b(?:HTTP\s*)?(?:401|403|404|408|429|500|502|503|504)\b`,
```
chandra447, line 83: **Act on:** drop bare HTTP `404` from this classifier.

Model/provider not-found already belongs on `retryWithoutOverrides` (subject ∧ reason). Treating a plain 404 as “try next `llmFallbackModels` entry” can burn paid attempts on non-model 404s, and routes model-404s through the fallback chain instead of strip-overrides.

Keep overload / rate-limit / 408 / 429 / 5xx / real transport signals. Auth failures (`401`/`403` / unauthorized language) are closer to the direct path’s auth-rejection bucket than transient transport — please don’t conflate them here either.

src/handlers/pi-child-process.ts
```diff
@@ -443,31 +490,55 @@ export async function execChildPrompt(
...
+        if (result.code === 0) return result;
+
+        const overrideResolutionFailure = shouldRetryWithoutOverrides(result);
+        if (hasFallback && (overrideResolutionFailure || shouldRetryWithFallback(result))) {
+          continue;
+        }
+        retryWithoutOverrides = !!options.retryWithoutOverrides
+          && hasChildLlmOverrides(config)
+          && overrideResolutionFailure;
+        if (!retryWithoutOverrides) return result;
```
chandra447, line 524: **Act on:** also check `hasPersistedProgress` on terminal failure paths (`return result` / `throw` when you are not continuing), especially the last configured model.

Today the shrink guard runs only before another paid attempt (`index > 0`) and before the strip-overrides retry. Call chain: last `pi.exec` → child tools already shrank disk → non-zero that does not `continue` → immediate failure return with no reload/shrink check → `triggerConsolidation` reports `consolidated: false` and the manual path only `loadFromDisk()` on success. That leaves the parent store stale vs disk and invites a later consolidator replay — the double-spend case this guard is meant to stop.

**Consider (same area):** encoding progress as synthetic `{ code: 0, stdout: "", stderr: "" }` overloads “child exited 0” with “disk already shrank.” Prefer an explicit skip discriminant, or lift the short-circuit to `triggerConsolidation`, so callers are not taught that empty `code: 0` means a successful child.

src/handlers/pi-child-process.ts
```diff
@@ -406,6 +432,27 @@ function shouldRetryWithoutOverridesForError(error: unknown): boolean {
   return shouldRetryWithoutOverridesFromText(String(error));
 }
 
+function shouldRetryWithFallbackFromText(text: string | undefined): boolean {
+  return !!text && RETRYABLE_PROVIDER_FAILURE.test(text);
+}
+
+function shouldRetryWithFallback(result: PiExecResult): boolean {
+  return shouldRetryWithFallbackFromText(result.stderr) || shouldRetryWithFallbackFromText(result.stdout);
```
chandra447, line 440: **Act on:** classify retryability from `stderr` (or structured child/provider metadata) only — not `stdout`.

The consolidation child often echoes prompt / memory / tool text on stdout. Failure memories commonly contain substrings this regex treats as transport failures (`overloaded`, `rate limit`, `503`, `timed out`, …). A non-retryable exit (e.g. `memory tool returned no changes`) can still schedule another paid model attempt if stdout happens to match.

Please add a regression test: matching corpus on stdout + non-retryable exit → no fallback attempt.

chandra447 (commented):
Thanks for the consolidation fallback work, @podledges — wiring `llmFallbackModels` through the `pi -p` / `pi.exec` path is the right fix for the gap left after #215. Design fits public Pi `ExtensionAPI` practices (still `pi.exec`, typed config, no new hosts).

(GitHub won’t let me mark REQUEST_CHANGES on a PR I opened from your branch, so this is a **COMMENT** review with the same bar: treat the three inline notes as **merge blockers**.)

**Before merge**
1. **Drop bare HTTP `404`** from `RETRYABLE_PROVIDER_FAILURE` — model-not-found stays on `retryWithoutOverrides`.
2. **Retry classifier: stderr only** (not stdout) — memory/prompt corpus on stdout can false-trigger paid fallbacks.
3. **`hasPersistedProgress` on terminal failure too** — last-model shrink must not report `consolidated: false` with a stale parent store.

**Consider (non-blocking)**
- Synthetic `{ code: 0 }` for “already shrunk” overloads `PiExecResult`; prefer an explicit skip discriminant or lift the short-circuit to `triggerConsolidation`.
- Split “next fallback model” from “strip overrides once” instead of OR-ing them in one loop.

**Noted:** large behavioral test coverage is good; a couple of small classifier/loop unit tests would still help. README `llmFallbackModels` row still leans Direct-Transport wording. `consolidationTimeoutMs` per-attempt docs look correct.

Happy to land a follow-up PR with 1–3 if that’s easier than pushing on this branch.

